PR #56, opened 2020-04-01
Make pod deletion synchronous

ldx:
This makes sure we don't keep pods in our registry after the user deletes them, thus they can be recreated right away in the same namespace with the same name.

I also added some extra checks to pod updates in the registry.

### DIFF
--- a/pkg/server/registry/pod_registry.go
+++ b/pkg/server/registry/pod_registry.go
@@ -212,6 +212,13 @@ func (reg *PodRegistry) UpdatePodSpecAndLabels(p *api.Pod) (*api.Pod, error) {
 		return nil, err
 	}
 	p, err = reg.AtomicUpdate(p.Name, func(in *api.Pod) error {
+		if api.IsTerminalPodPhase(in.Spec.Phase) {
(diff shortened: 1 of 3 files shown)


pkg/server/registry/pod_registry.go
```diff
@@ -212,6 +212,13 @@ func (reg *PodRegistry) UpdatePodSpecAndLabels(p *api.Pod) (*api.Pod, error) {
 		return nil, err
 	}
 	p, err = reg.AtomicUpdate(p.Name, func(in *api.Pod) error {
+		if api.IsTerminalPodPhase(in.Spec.Phase) {
+			return fmt.Errorf("pod is in a terminal pod phase")
```
justnoise, line 216: one small change: can we have this say 'pod **spec** is in a terminal pod phase".  It's a minor difference but a bit clearer since we have 2 phases and the wording "is in a terminal pod phase" makes it sound like it's referring to the status of the pod

ldx: Sounds good :+1:

PR #61, opened 2020-04-09
Add annotation for creating pod cloud routes

ldx:
Changes:
- New `pod.elotl.co/cloud-route: <cidr1 cidr2 ...>` annotation, that will make Kip add a route in the cloud subnet route table, using the instance backing the pod as the next hop for the CIDRs.
- I refactored the setup for DNS and how the network agent token is created, to reduce clutter in the pod controller, and injecting them as PodController dependencies.
- I threaded through the `--cluster-domain` parameter from node-cli, so the cluster domain can now be overridden.
- Added a `--cluster-dns` parameter to make the DNS used by pods with a `ClusterFirst` DNS policy configurable.

The main use case right now is to allow a caching DNS server in the VPC subnet for a hybrid on-prem/cloud setup: the caching DNS pod can add a route to its extra IP address (e.g. node-local-dns has a link-local listener by default), and the operator can also inject this IP via the `--cluster-dns <same link-local address>` parameter to Kip.

### DIFF
--- a/pkg/server/cloud/aws/network.go
+++ b/pkg/server/cloud/aws/network.go
@@ -331,11 +331,6 @@ func (e *AwsEC2) ModifySourceDestinationCheck(instanceID string, isEnabled bool)
 }
 
 func (e *AwsEC2) RemoveRoute(destinationCIDR, instanceID string) error {
-	if destinationCIDR == "" || instanceID == "" {
-		return fmt.Errorf(
-			"invalid input: at least one non-empty value needed (got %q %q)",
-			destinationCIDR, instanceID)
-	}
 	out, err := e.client.DescribeRouteTables(&ec2.DescribeRouteTablesInput{
 		Filters: []*ec2.Filter{
 			{
@@ -360,13 +355,24 @@ func (e *AwsEC2) RemoveRoute(destinationCIDR, instanceID string) error {
 			continue
 		}
 		for _, route := range table.Routes {
+			// Ignore any routes that are not instance routes.
+			routeIID := aws.StringValue(route.InstanceId)
+			if routeIID == "" {
+				continue
+			}
+			// If instanceID is provided, match on it.
+			if instanceID != "" && instanceID != routeIID {
+				continue
+			}
+			// If destinationCIDR is provided, match on it.
 			if destinationCIDR != "" &&
 				destinationCIDR != aws.StringValue(route.DestinationCidrBlock) {
 				continue
 			}
-			// Ignore any routes that are not instance routes.
-			routeIID := aws.StringValue(route.InstanceId)
-			if routeIID == "" || instanceID != "" && instanceID != routeIID {
+			// If both instanceID and destinationCIDR are empty, remove
+			// blackhole instance routes.
+			if destinationCIDR == "" && instanceID == "" &&
+				aws.StringValue(route.State) != ec2.RouteStateBlackhole {
 				continue
 			}
 			_, err = e.client.DeleteRoute(&ec2.DeleteRouteInput{
--- a/pkg/server/cloud/aws/network.go
+++ b/pkg/server/cloud/aws/network.go
@@ -330,57 +330,85 @@ func (e *AwsEC2) ModifySourceDestinationCheck(instanceID string, isEnabled bool)
 	return err
 }
 
-func (e *AwsEC2) RemoveRoute(destinationCIDR string) error {
+func (e *AwsEC2) RemoveRoute(destinationCIDR, instanceID string) error {
+	if destinationCIDR == "" || instanceID == "" {
+		return fmt.Errorf(
+			"invalid input: at least one non-empty value needed (got %q %q)",
+			destinationCIDR, instanceID)
+	}
 	out, err := e.client.DescribeRouteTables(&ec2.DescribeRouteTablesInput{
 		Filters: []*ec2.Filter{
 			{
-				Name: aws.String("vpc-id"),
-				Values: []*string{
-					aws.String(e.vpcID),
-				},
+				Name:   aws.String("vpc-id"),
+				Values: aws.StringSlice([]string{e.vpcID}),
+			},
+			{
+				Name:   aws.String("association.main"),
+				Values: aws.StringSlice([]string{"false"}),
+			},
+			{
+				Name:   aws.String("association.subnet-id"),
+				Values: aws.StringSlice([]string{e.subnetID}),
 			},
 		},
 	})
 	if err != nil {
-		return util.WrapError(err, "Could not list vpc route tables")
+		return util.WrapError(err, "listing vpc route tables")
 	}
 	for _, table := range out.RouteTables {
 		if len(table.Associations) == 0 {
 			continue
 		}
 		for _, route := range table.Routes {
-			if destinationCIDR == aws.StringValue(route.DestinationCidrBlock) {
-				_, err = e.client.DeleteRoute(&ec2.DeleteRouteInput{
-					DestinationCidrBlock: route.DestinationCidrBlock,
-					RouteTableId:         table.RouteTableId,
-				})
-				if err != nil {
-					return util.WrapError(
-						err,
-						"Error deleting old route table entry for table %s",
-						*table.RouteTableId,
-					)
-				}
-				break
+			if destinationCIDR != "" &&
+				destinationCIDR != aws.StringValue(route.DestinationCidrBlock) {
+				continue
 			}
+			// Ignore any routes that are not instance routes.
+			routeIID := aws.StringValue(route.InstanceId)
+			if routeIID == "" || instanceID != "" && instanceID != routeIID {
+				continue
+			}
+			_, err = e.client.DeleteRoute(&ec2.DeleteRouteInput{
+				DestinationCidrBlock: route.DestinationCidrBlock,
+				RouteTableId:         table.RouteTableId,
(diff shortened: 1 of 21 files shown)


pkg/server/cloud/aws/network.go
```diff
@@ -316,57 +316,85 @@ func (e *AwsEC2) ModifySourceDestinationCheck(instanceID string, isEnabled bool)
...
-					)
-				}
-				break
+			if destinationCIDR != "" &&
+				destinationCIDR != aws.StringValue(route.DestinationCidrBlock) {
+				continue
+			}
+			// Ignore any routes that are not instance routes.
+			routeIID := aws.StringValue(route.InstanceId)
+			if routeIID == "" || instanceID != "" && instanceID != routeIID {
```
justnoise, line 355: This is kinda confusing.  I don't think you need the `instanceID != "" && ` part of this expression.  Just `routeIID == "" || instanceID != routeIID`.  Can you double check this? I might be wrong...

1. We've checked that instanceID != "" at the top of the function
2. If `instanceID == ""` then the check `instanceID != routeID` will evaluate to false only if `routeID == ""`.  That's already checked.

pkg/server/pod_controller.go
```diff
@@ -491,6 +449,29 @@ func (c *PodController) dispatchPodToNode(pod *api.Pod, node *api.Node) {
...
 }
 
+func (c *PodController) addCloudRoute(node *api.Node, cidrs []string) error {
+	instanceID := node.Status.InstanceID
+	err := c.cloudClient.ModifySourceDestinationCheck(instanceID, false)
+	if err != nil {
+		return err
+	}
+	for _, cidr := range cidrs {
+		if _, _, err := net.ParseCIDR(cidr); err != nil {
```
justnoise, line 459: Can we check that this parses in convert.go or api/validation/validation.go?  It might be better to ensure things are in order before accepting the pod.

pkg/server/cloud/azure/network.go
```diff
@@ -336,6 +361,11 @@ func (az *AzureClient) RemoveRoute(destinationCIDR string) error {
 }
 
 func (az *AzureClient) AddRoute(destinationCIDR, instanceID string) error {
+	if destinationCIDR == "" && instanceID == "" {
```
justnoise, line 364: This check is an `||` in the aws code.  Curious if you meant `||`

pkg/server/cloud/aws/network.go
```diff
@@ -316,57 +316,85 @@ func (e *AwsEC2) ModifySourceDestinationCheck(instanceID string, isEnabled bool)
...
+				Name:   aws.String("association.subnet-id"),
+				Values: aws.StringSlice([]string{e.subnetID}),
 			},
 		},
 	})
 	if err != nil {
-		return util.WrapError(err, "Could not list vpc route tables")
+		return util.WrapError(err, "listing vpc route tables")
 	}
 	for _, table := range out.RouteTables {
```
justnoise, line 399: Any idea what happens if we try to add 2 identical destinationCIDRs to a route table?  If that causes an error, one thing we could do is catch that error, delete the old route and then try to re-add the route.

If it doesn't cause an error: :tada:

pkg/server/pod_controller.go
```diff
@@ -491,6 +449,29 @@ func (c *PodController) dispatchPodToNode(pod *api.Pod, node *api.Node) {
...
+			return err
+		}
+		if err := c.cloudClient.AddRoute(cidr, instanceID); err != nil {
```
justnoise, line 462: GarbageController doesn't clean up routes yet.  Should that get added?

pkg/server/cloud/azure/network.go
justnoise: Might want to check the `||`s and `&&`s in front of all the `AddRoute` and `DeleteRoute` implementations.

pkg/server/cloud/aws/network.go
justnoise: I think this is answered in a comment later on.  I'm curious how we should handle this. Can we delete the old route?  The alternative is to keep churning instances for this pod.

pkg/server/pod_controller.go
justnoise: Humm... I just found a need to validate an annotation as well and there wasn't scaffolding for doing that in api/validation/validate.go.  Here's the function I created:

```go
func ValidatePodAnnotations(annotations map[string]string, fldPath *field.Path) field.ErrorList {
	allErrs := field.ErrorList{}
	for k, v := range annotations {
		switch k {
		case apiannotations.PodResourcesPrivateIPOnly:
			_, err := strconv.ParseBool(v)
			fmt.Println("parsed bool")
			if err != nil {
				allErrs = append(allErrs, field.Invalid(fldPath.Child(k), v, "Could not parse annotation value as bool"))
			}
		}
	}
	return allErrs
}
```

Called in `ValidatePod`:

```go
allErrs = append(allErrs, ValidatePodAnnotations(pod.Annotations, field.NewPath("metadata.annotations"))...)
```

pkg/server/nodemanager/node_controller.go
```diff
@@ -330,6 +330,7 @@ func (c *NodeController) stopSingleNode(node *api.Node) error {
 	}
 	c.NodeClientFactory.DeleteClient(node.Status.Addresses)
 	go func(n *api.Node) {
+		_ = c.CloudClient.RemoveRoute("", n.Status.InstanceID)
```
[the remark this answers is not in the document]
ldx: > In the current implementation, this results in a noop in the cloudClient (and returns an error)? Seems like the check in AWS is if cidr == "" || instanceID == "" { return fmt.Errorf(...)}

Thanks, that was a bug. In RemoveRoute(), we need at least one input parameter (either the instance ID or the CIDR, and RemoveRoute() will match for one if only one of them is non-empty, otherwise for both), in AddRoute() we need both.

> I'm unsure if this is needed here since we're also removing the route in the pod controller. 

I added it to both places, since there are edge cases around Kip getting restarted when the pod is being terminated. I think 99.99% of the time it is unnecessary to call RemoveRoute() from both places.

Let me remove the second call, and instead clean up dangling instance routes via the garbage controller. We can simply check blackhole instance routes and remove them there instead of the extra RemoveRoute() call from the node controller.

pkg/server/cloud/aws/network.go
```diff
@@ -316,57 +316,85 @@ func (e *AwsEC2) ModifySourceDestinationCheck(instanceID string, isEnabled bool)
...
-				})
-				if err != nil {
-					return util.WrapError(
-						err,
-						"Error deleting old route table entry for table %s",
-						*table.RouteTableId,
-					)
-				}
-				break
+			if destinationCIDR != "" &&
```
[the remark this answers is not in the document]
ldx: RemoveRoute() now can handle 4 scenarios:
- destinationCIDR and instanceID are both set, it will remove routes that point to the instance for the prefix in destination CIDR.
- destinationCIDR empty, instanceID set: it will remove routes (any prefix) that point to the instance.
- destinationCIDR set, instanceID empty: it will remove routes for the prefix in destinationCIDR, if it finds an instance route (for any instance).
- destinationCIDR and instanceID are both empty: it will remove blackhole instance routes (i.e. any inactive route pointing to an instance).

So destinationCIDR can be empty here.

pkg/server/cloud/aws/network.go
ldx: I removed the check at the top of the function, since we are now cleaning up dangling instance routes via leaving both input variables empty.

pkg/server/pod_controller.go
ldx: :+1:

pkg/server/cloud/azure/network.go
ldx: Thanks, I fixed them.

pkg/server/cloud/aws/network.go
ldx: I'd rather not delete an existing route. If the operator accidentally tries to add the same prefix via multiple pods, one will succeed, the others will fail. The cause of the failure will pop up via logs and the pod status. If we silently remove existing active routes, then things will *seem* to work (no imminent pod failure), except likely routing will be misconfigured.

ldx: BTW this is why I added extra calls to RemoveRoute(): if a pod is terminated, we should ensure that eventually its cloud route is also gone.

pkg/server/pod_controller.go
ldx: Adding it.

pkg/server/nodemanager/node_controller.go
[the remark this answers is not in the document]
justnoise: Cleaning up blackhole routes sounds good.

pkg/server/cloud/aws/network.go
```diff
@@ -316,57 +316,97 @@ func (e *AwsEC2) ModifySourceDestinationCheck(instanceID string, isEnabled bool)
 	return err
 }
 
-func (e *AwsEC2) RemoveRoute(destinationCIDR string) error {
+func (e *AwsEC2) RemoveRoute(destinationCIDR, instanceID string) error {
```
justnoise, line 319: I like how this function turned out!  It's easy to understand what's going on in multiple cases.

PR #63, opened 2020-04-13
Fix host port removal

ldx:
When a pod is being terminated, the event is sent out while the pod is
in the 'Running' phase in the registry (unless it has never started up
correctly). Change event handling in server.go to be able to correctly
remove host ports from the PodSpec.


justnoise:
This looks correct.  The logic around pod termination is unfortunately twisty and poorly named.  I think that could eventually be improved with a bit of tracing through everything.

:+1:

PR #86, opened 2020-05-15
GCE

justnoise:


pkg/server/cloud/gce/instances.go
```diff
@@ -0,0 +1,445 @@
...
+	return diskSpec
+}
+
+func (c *gceClient) getInstanceNetworkSpec(privateIPOnly bool) []*compute.NetworkInterface {
+	networkURL := c.getNetworkURL()
+	subNetworkURL := c.getSubNetworkURL()
+
+	accessConfig := &compute.AccessConfig{
+		Name:        "External NAT",
+		NetworkTier: "STANDARD",
```
ldx, line 116: [Not a blocker, let's add an issue for later?] I think we should make this configurable in provider.yaml, and even allow overriding it via an annotation. The price difference is not that significant: https://cloud.google.com/network-tiers/pricing but the premium tier sounds good and desirable for certain customers.

pkg/server/cloud/gce/instances.go
```diff
@@ -0,0 +1,445 @@
...
+		Type:        "ONE_TO_ONE_NAT",
+	}
+	// If we are only using private IP set the access config to nil
+	if privateIPOnly || !c.usePublicIPs {
+		accessConfig = nil
```
ldx, line 121: Correct me if I'm wrong: even with privateIPOnly, cells should still be able to access the internet, shouldn't they?

pkg/server/cloud/gce/instances.go
```diff
@@ -0,0 +1,445 @@
...
+	return spec, nil
+}
+
+func (c *gceClient) StartNode(node *api.Node, metadata string) (*cloud.StartNodeResult, error) {
+	klog.V(2).Infof("Starting instance for node: %v", node)
+	spec, err := c.createInstanceSpec(node, metadata)
+	if err != nil {
+		return nil, err
+	}
+	klog.V(2).Infof("Starting node with security groups: %v subnet: '%s'",
```
ldx, line 186: Tip: you can use the `%q` formatter for printing quoted strings.

pkg/server/cloud/gce/instances.go
```diff
@@ -0,0 +1,445 @@
...
+	// see pkg/server/cloud/aws/instances.StartNode()
+	if err := c.waitOnOperation(op.Name, c.getZoneOperation); err != nil {
+		return nil, err
+	}
+	startResult := &cloud.StartNodeResult{
+		InstanceID:       spec.Name,
+		AvailabilityZone: c.zone,
+	}
+	return startResult, nil
+}
```
ldx, line 228: Is this function a copy/paste of StartNode()? Wouldn't it be easier to call it while spot node support is not implemented?

pkg/server/cloud/gce/instances.go
```diff
@@ -0,0 +1,445 @@
...
+		if err != nil {
+			klog.Errorf("Error waiting for instance to start: %v", err)
+			return nil, err
+		}
+
+		klog.V(5).Infof("status: %s", status)
+		if status == statusInstanceRunning {
+			break
+		}
+		time.Sleep(10 * time.Second)
```
ldx, line 242: Two things:
- 10s is probably a bit long - can we reduce it to something like 5s? With 10s we might wait for an extra 10s if the instance comes up right after the last instance status query.
- Shouldn't we add a timeout? What happens when we start the instance, but it crashes/terminated outside of Kip, and never gets to running? Will we get stuck in this loop?

pkg/server/cloud/gce/instances.go
```diff
@@ -0,0 +1,445 @@
...
+	defer cancel()
+	resp, err := c.service.Images.Get(project, image).Context(ctx).Do()
+	if err != nil {
+		return "", util.WrapError(err, "Error looking up boot image %s/%s", project, image)
+	}
+	if resp == nil {
+		return "", nilResponseError("Images.Get")
+	}
+	return resp.SelfLink, nil
+}
```
ldx, line 440: This will need a simple change since I've change the cloud API GetImageID() -> GetImage() to save cloud-specific image attributes that are needed when starting a new instance (we use it on AWS to save the disk device used by the image).

pkg/server/cloud/gce/instances.go
```diff
@@ -0,0 +1,445 @@
...
+
+func (c *gceClient) StartNode(node *api.Node, metadata string) (*cloud.StartNodeResult, error) {
+	klog.V(2).Infof("Starting instance for node: %v", node)
+	spec, err := c.createInstanceSpec(node, metadata)
+	if err != nil {
+		return nil, err
+	}
+	klog.V(2).Infof("Starting node with security groups: %v subnet: '%s'",
+		c.bootSecurityGroupIDs, c.subnetName)
+	op, err := c.service.Instances.Insert(c.projectID, c.zone, spec).Do()
```
ldx, line 188: It seems a context is used for timeouts everywhere else, any reason not to do the same here?

pkg/server/cloud/gce/instances.go
```diff
@@ -0,0 +1,445 @@
...
+
+func (c *gceClient) StartSpotNode(node *api.Node, metadata string) (*cloud.StartNodeResult, error) {
+	klog.V(2).Infof("Starting instance for node: %v", node)
+	spec, err := c.createInstanceSpec(node, metadata)
+	if err != nil {
+		return nil, err
+	}
+	klog.V(2).Infof("Starting node with security groups: %v subnet: '%s'",
+		c.bootSecurityGroupIDs, c.subnetName)
+	op, err := c.service.Instances.Insert(c.projectID, c.zone, spec).Do()
```
ldx, line 213: Here too, how about using a context for timeouts?

pkg/server/cloud/gce/instances.go
```diff
@@ -0,0 +1,445 @@
...
+		klog.Errorf("Error retrieving Pod IP: %v", err)
+		return nil, err
+	}
+	addresses = api.SetPodIP(podIP, addresses)
+
+	return addresses, nil
+}
+
+func (c *gceClient) StopInstance(instanceID string) error {
+	op, err := c.service.Instances.Delete(c.projectID, c.zone, instanceID).Do()
```
ldx, line 285: Context for timeout?

pkg/server/cloud/gce/instances.go
jrroman: Yes, I agree there are a few of these I want to add as annotations, another one is the `SATA` vs `NVMe` for disk interface type (line 98 instances.go)

pkg/server/cloud/gce/instances.go
jrroman: resolved in commit 2924d18

pkg/server/cloud/gce/firewall.go
```diff
@@ -0,0 +1,304 @@
...
+func gceProtocolToKipProtocol(proto string) api.Protocol {
+	proto = strings.ToLower(proto)
+	switch proto {
+	case "tcp":
+		return api.ProtocolTCP
+	case "udp":
+		return api.ProtocolUDP
+	case "sctp":
+		return api.ProtocolSCTP
+	case "ICMP":
```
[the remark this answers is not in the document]
jrroman: commit b1ff36e resolved this

pkg/server/cloud/gce/firewall.go
```diff
@@ -0,0 +1,304 @@
...
+	return nil
+}
+
+func (c *gceClient) FindSecurityGroup(sgName string) (*cloud.SecurityGroup, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
+	defer cancel()
+	resp, err := c.service.Firewalls.Get(c.projectID, sgName).Context(ctx).Do()
+	if err != nil {
+		if isNotFoundError(err) {
+			return nil, nil
```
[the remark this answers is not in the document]
jrroman: commit aef59c3 fixes this

pkg/server/cloud/gce/instances.go
jrroman: fixed in d7bc5a7

pkg/server/cloud/gce/instances.go
jrroman: fixed in d7bc5a7

pkg/server/cloud/gce/instances.go
jrroman: fixed in d7bc5a7

pkg/server/cloud/gce/instances.go
jrroman: 1. I have decreased sleep time to 5 seconds in d7bc5a7
2. I believe @justnoise  is going to take a look at the timeout situation.

pkg/server/cloud/gce/firewall.go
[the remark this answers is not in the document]
justnoise: The callers of this code assume that FindSecurityGroup will not return an error if the security group is not found.  Is it an error for a find operation to not find anything?  In AWS it is not an error, you just get back a list of security groups with len == 0.  We've gone with that convention in GCE as well.  I went with that behavior since its actually easier to deal with that vs. checking error types.

Personally, I'd like to keep things as they were but the question is where we want consistency. Long ago, FindSecurityGroup was in the CloudClient interface so we'd want to stay consistent with the implementation in AWS.  It's no longer in the interface...  If you'd really like for this to return an error and you think it should behave differently than AWS when a security group isn't found, please update callers of this as well to handle the error appropriately.  By returning an error, callers of this fuction now break.

pkg/server/cloud/gce/instances.go
justnoise: No.  If you don't have an external address your traffic is dropped at the edge of the VPC. You must have some type of instance proxy or Cloud NAT.

* https://cloud.google.com/vpc/docs/vpc#internet_access_reqs
* https://cloud.google.com/vpc/docs/advanced-vpc#detailedconnectionwalkthrough

pkg/server/cloud/gce/instances.go
```diff
@@ -0,0 +1,456 @@
...
+// In we dictate whether the node is a spot based on the node passed in
+// this is decided in createInstanceSpec which is called in the unexported
+// startNode function. StartSpotNode is necessary to fullfil the interface.
+func (c *gceClient) StartSpotNode(node *api.Node, metadata string) (*cloud.StartNodeResult, error) {
+	return c.startNode(node, metadata)
+}
+
+func (c *gceClient) WaitForRunning(node *api.Node) ([]api.NetworkAddress, error) {
+	for {
+		status, err := c.getInstanceStatus(node.Status.InstanceID)
```
justnoise, line 241: The call right before this is StartNode which is a write operation to the cloud.  Those writes are eventually consistent.  That means the instanceID might not _really_ exist  in GCE when we make the call to getInstanceStatus.  I'll work on figuring out the retry logic for this one.

1. What does it look like when the instance doesn't exist yet.  Can we handle that for 30s.  After 30s, if the instance doesn't exist, we should return an error.
2. We need to time out the waitForRunning operation so this eventually ends if the instance never comes up, gets deleted behind the scenes, etc.

pkg/server/cloud/gce/instances.go
jrroman: made updates for this in 8fe2c2c

pkg/server/cloud/gce/firewall.go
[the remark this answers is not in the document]
jrroman: Understood I actually just experienced this now. I have since reverted this issue, thank you for pointing this out @justnoise

pkg/server/cloud/gce/instances.go
justnoise: Fixed this up in https://github.com/elotl/kip/commit/b4f0e5cd7bc7a7ccbb3ebc3ca0b53ecb22fd1e19

pkg/server/cloud/gce/instances.go
justnoise: https://github.com/elotl/kip/commit/b4f0e5cd7bc7a7ccbb3ebc3ca0b53ecb22fd1e19

PR #87, opened 2020-05-15
Add GCP Terraform config

ldx:
This is a simple and minimalistic TF config example, using GKE.

### DIFF
--- a/deploy/terraform-gcp/main.tf
+++ b/deploy/terraform-gcp/main.tf
@@ -1,5 +1,5 @@
 provider "google" {
-  project = "excellent-ion-277200"
+  project = var.project
   region  = var.region
   zone    = var.zone
 }
--- a/deploy/terraform-gcp/main.tf
+++ b/deploy/terraform-gcp/main.tf
@@ -146,6 +146,8 @@ resource "null_resource" "client_permissions" {
 }
 
 resource "null_resource" "deploy" {
+  count = length(var.kustomize-dir) > 0 ? 1 : 0
+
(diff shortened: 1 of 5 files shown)


deploy/terraform-gcp/main.tf
```diff
@@ -0,0 +1,168 @@
+provider "google" {
+  project = "excellent-ion-277200"
```
jrroman, line 2: I believe the project will have to be filled in by the user or set by an env for their specific project in gcp.

deploy/terraform-gcp/main.tf
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    oauth_scopes = [
+      "https://www.googleapis.com/auth/logging.write",
+      "https://www.googleapis.com/auth/monitoring",
+    ]
+  }
+}
+
+resource "google_filestore_instance" "filestore" {
```
justnoise, line 71: Do we want to include filestore along with this manifest?  I was under the impression that would be an enterprisey add-on.

deploy/terraform-gcp/main.tf
ldx: Thanks, that's a leftover from my test setup. I'll fix it.

deploy/terraform-gcp/main.tf
ldx: Yes, this is just a few lines, nothing fancy. The business logic for image caching will be in a separate controller and itzo add-on.

Issue #70, opened 2020-04-27
Generate Kubelet API certificate for Kip


ldx:
Right now we mount in the cert from the kubelet Kip runs on (from /var/lib/kubelet/pki). However, this is very inelegant and not something that should be used in production.

Instead, we could generate a certificate for Kip ourselves, either in an init container, or in Kip when it starts up and there's no certificate. The certificate could then be stored on the persistent volume. See https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/server.go#L978-L993 how the kubelet generates its own certificate when it's missing at startup.

Note: this is not the client cert kubelets use to communicate with the API server as their client cert. Kip uses its service account token for that purpose.

ldx:
@myechuri for now, we can make the symlinks work via mounting it into the right directory on VK:

        env:
        - name: APISERVER_CERT_LOCATION
          value: /var/lib/kubelet/pki/kubelet-client-current.pem
        - name: APISERVER_KEY_LOCATION
          value: /var/lib/kubelet/pki/kubelet-client-current.pem

        volumeMounts:
        - mountPath: /var/lib/kubelet/pki
          name: kubelet-pki

      volumes:
      - hostPath:
          path: /var/lib/kubelet/pki
          type: Directory
        name: kubelet-pki

ldx:
This has been fixed via d99177fa

PR #88, opened 2020-05-16
Generate server certificate in an init container

ldx:
I added a simple init container to VK that generates a server certificate for serving the API. Unfortunately, we have to jump through a few extra hoops (approving the CSR ourselves) since VK uses its own service account for interacting with the control plane.

I made sure the certificate is only generated if it's missing or invalid.

### DIFF
--- a/scripts/init-cert/Dockerfile
+++ b/scripts/init-cert/Dockerfile
@@ -1,0 +1,20 @@
+FROM debian:stable-slim
+
+ENV DEBIAN_FRONTEND "noninteractive"
+
+RUN apt-get update -y && \
+        apt-get dist-upgrade -y && \
+        apt-get install -y curl gettext-base iproute2 openssl
+
+RUN mkdir -p /usr/local/bin
+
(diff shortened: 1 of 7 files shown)


ldx:
This fixes #70

scripts/init-cert/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN curl -fL https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl > /usr/local/bin/kubectl
+RUN chmod +x /usr/local/bin/kubectl
+
+RUN curl -fL https://github.com/ldx/token2kubeconfig/releases/download/v0.0.2/token2kubeconfig-amd64 > /usr/local/bin/token2kubeconfig
+RUN chmod +x /usr/local/bin/token2kubeconfig
+
+RUN mkdir -p /opt/csr
+COPY csr /opt/csr
+
+CMD /opt/csr/create.sh
```
justnoise, line 20: should this be create-csr.sh?

ldx: Correct, thanks!

PR #89, opened 2020-05-20
Add more debug logging when creating provider

ldx:
Two changes:
- I added some V(5) logs when the provider is being created.
- In Makefile, one can now use the `DKR` environment variable to choose `img`, `docker` or any other command line compatible client for building and pushing the kip image, e.g. `make img DKR=docker` will build the image via docker.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+DKR=img
 GIT_VERSION=$(shell git describe --dirty)
 CURRENT_TIME=$(shell date +%Y%m%d%H%M%S)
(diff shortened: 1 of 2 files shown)


ldx:
> Should we be updating
> 
> https://github.com/elotl/kip/blob/c0c246b61a6885f20ab74e42aca2f85f2caaa5dc/deploy/manifests/virtual-kubelet/base/deployment.yaml#L71
> to use `v=5`?
> 
> Other than that LGTM +1

I think that's fine for the default deployment. Users can easily override it via on overlay.

Issue #96, opened 2020-05-20
Support dockerconfigjson registry secrets


justnoise:
In k8s, the standard way to encode registry secrets is:
```
kubectl create secret docker-registry regcred --docker-server=<server> --docker-username=<username> --docker-password=<password> --docker-email=<docker-email>
```

This produces a secret that looks like:

```yaml
apiVersion: v1
kind: Secret
data:
  .dockerconfigjson: <base64 encoded json>
```

The decoded json looks like:

```javascript
{
    "auths": {
        "docker.io": {
            "username": "<username>",
            "password": "<password>",
            "email": "<docker-email>",
            "auth": "<base64 encoded username:password>"
        }
    }
}
```
Kip pulls out image secrets and sends them to cells in a structure containing {server, username, password}.  We should try to pull that data out of a docker formatted secret.

Lets update kip to correctly pull out the necessary data from a docerconfigjson secret.

PR #98, opened 2020-05-21
fix up using imagePullSecrets in kip, be able to use dockerconfigjson

justnoise:
K8s docs for image pull specs are here: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/

This is a first pass at being able to use image pull secrets created with kubectl.  It's not perfect (I think docker's config.json allows you to specify different creds for the same repo URL at different paths) but it's a decent start.

Creating docker secrets:
```bash
$ kubectl create secret docker-registry regcred \
    --docker-server=<server> \
    --docker-username=<username> \
    --docker-password=<password> \
    --docker-email=<docker-email>

$kubectl create secret generic file-regcred \ 
    --from-file=.dockerconfigjson=/home/myuser/.docker/config.json \
    --type=kubernetes.io/dockerconfigjson
```

The first case creates a secret detailed here: https://github.com/elotl/kip/issues/96
The second case creates a secret with the entire contents of /home/myuser/.docker/config.json base64 encoded.  For that case, the docker server can either be a hostname (e.g. `689494258501.dkr.ecr.us-east-1.amazonaws.com`) or a full url (`https://index.docker.io/v1/`).

This PR tries to understand both formats do work to extract the correct server credentials, either from the username and password fields or falling back to the auth field.  It's possible we could just use the auth field...  I'm unsure if we should be defensive or make that simplification.

### DIFF
--- a/pkg/server/pod_controller.go
+++ b/pkg/server/pod_controller.go
@@ -262,12 +262,7 @@ func parseDockerConfigServer(serverURL string) (string, error) {
 }
 
 const (
-	dockerServerKey       = "server"
-	dockerUsernameKey     = "username"
-	dockerPasswordKey     = "password"
-	dockerAuthsKey        = "auths"
-	dockerConcatedAuthKey = "auth"
-	dockerConfigJSONKey   = ".dockerconfigjson"
+	dockerConfigJSONKey = ".dockerconfigjson"
 )
 
 func parseDockerConfigCreds(dockerJSON []byte) (map[string]api.RegistryCredentials, error) {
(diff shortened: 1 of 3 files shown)


pkg/server/pod_controller.go
```diff
@@ -247,36 +249,119 @@ func (c *PodController) markFailedPod(pod *api.Pod, startFailure bool, msg strin
...
+	//             "auth": "base64(username:password)"
+	//         }
+	//         "689494258501.dkr.ecr.us-east-1.amazonaws.com": {
+	//              "auth": "base64(username:password)"
+	//         }
+	//     }
+	// }
+	type dockerConfig struct {
+		Auths map[string]map[string]string `json:"auths"`
+	}
```
ldx, line 290: It's a simple data structure, but in case you want use stricter typing: https://github.com/docker/cli/blob/master/cli/config/types/authconfig.go

pkg/server/pod_controller_test.go
```diff
@@ -495,3 +495,91 @@ func TestSetPodDispatchingParams(t *testing.T) {
...
+		{
+			config: []byte(`{
+    "auths": {
+        "https://index.docker.io/v1/": {
+            "auth": "this*@aint--base_64"
+        }
+    }
+}`),
+			err: true,
+		},
```
ldx, line 556: Are these two test cases the same?

pkg/server/pod_controller.go
justnoise: That works.

PR #99, opened 2020-05-22
Handle broken metadata response in google cloud

justnoise:
For some reason, when you curl the gce metadata service for the zone, from alpine, an extra slash gets added to the response:

```bash
exec -it curler -- /bin/ash
# curl "http://metadata.google.internal/computeMetadata/v1/instance/zone" -H "Metadata-Flavor: Google"
projects/832569367454/zones/us-west1-a/
```

This does not happen in other distros (debian and ubuntu were tested)
```
kk exec -it curler -- curl "http://metadata.google.internal/computeMetadata/v1/instance/zone" -H "Metadata-Flavor: Google"
projects/832569367454/zones/us-west1-a
```

The library we use assumes the response doesn't end with a slash and it'll break if it does end in a slash.  This PR gets the zone from the metadata ourselves and handle both cases.

### DIFF
--- a/pkg/server/cloud/gce/metadata.go
+++ b/pkg/server/cloud/gce/metadata.go
@@ -17,6 +17,7 @@ limitations under the License.
 package gce
 
 import (
+	"fmt"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


ldx:
Does this only happen when you use curl, or in kip too? I think the default curl in alpine is just a busybox alias, you need to add curl explicitly via apk to get the "real" one.

justnoise:
I noticed this happening inside kip first.  It would return emptystring for the zone which would mess up autodetecting its environment.  I then investigated on alpine by apk adding curl and testing it out. (busybox doesn't come with its own curl but it does implement wget)  The really strange thing is that the problem doesn't show up when you `exec virtual-kubelet -- curl <metadata URL>` whithout the interactive shell...  

Overall, I'm not sure what the heck is causing the problem but we can easily work around it.

PR #103, opened 2020-05-25
Update stats

ldx:
This adds more metrics to the stats summary API, and makes Kip use the default metrics port by default. The default port is pretty much a requirement when using metrics-server; even though the port is configurable, it only is on a per-cluster (not per-node) basis. That is, if we want to collect metrics from nodes we have to choose one port all kubelet must use for the stats summary endpoint.

To fix this, I made Kip NOT use host network mode by default. I added kube-proxy in a sidecar for setting up NodePorts in the VK pod.

Since the IP address now will likely change after every pod restart, I also added a check to the init container when generating the certificate, so that the cert will be re-created if the IP has changed.

With these changes, the stats are a lot more detailed:

      {
         "containers" : [
            {
               "cpu" : {
                  "time" : "2020-05-26T22:09:42Z",
                  "usageCoreNanoSeconds" : 228726035073,
                  "usageNanoCores" : 258544883
               },
               "memory" : {
                  "majorPageFaults" : 53,
                  "rssBytes" : 7199850496,
                  "time" : "2020-05-26T22:09:42Z",
                  "usageBytes" : 8749936640,
                  "workingSetBytes" : 8109391872
               },
               "name" : "buildkite",
               "startTime" : "2020-05-26T22:00:57Z"
            }
         ],
         "cpu" : {
            "time" : "2020-05-26T22:09:42Z",
            "usageCoreNanoSeconds" : 228726035073,
            "usageNanoCores" : 258544883
         },
         "ephemeral-storage" : {
            "availableBytes" : 7408414720,
            "capacityBytes" : 10222866432,
            "inodes" : 1290240,
            "inodesFree" : 1165646,
            "inodesUsed" : 124594,
            "time" : "2020-05-26T22:09:42Z",
            "usedBytes" : 2797674496
         },
         "memory" : {
            "majorPageFaults" : 53,
            "rssBytes" : 7199850496,
            "time" : "2020-05-26T22:09:42Z",
            "usageBytes" : 8749936640,
            "workingSetBytes" : 8109391872
         },
         "network" : {
            "interfaces" : [
               {
                  "name" : "eth0",
                  "rxBytes" : 716386029,
                  "txBytes" : 8983003
               }
            ],
            "name" : "eth0",
            "rxBytes" : 716386029,
            "time" : "2020-05-26T22:09:42Z",
            "txBytes" : 8983003
         },
         "podRef" : {
            "name" : "android-ndk-f49b9496c-nlxsr",
            "namespace" : "default",
            "uid" : "1c6c2b41-9f9a-11ea-8636-42010a800163"
         },
         "startTime" : "2020-05-26T22:00:57Z"
    }

The GKE monitoring dashboard also shows pod and container metrics from VK workloads:

image

image

Related itzo PR: https://github.com/elotl/itzo/pull/88

### DIFF
--- a/deploy/manifests/virtual-kubelet/base/deployment.yaml
+++ b/deploy/manifests/virtual-kubelet/base/deployment.yaml
@@ -69,6 +69,7 @@ spec:
         - --disable-taint
         - --klog.logtostderr
         - --klog.v=2
+        - --metrics-addr=:10255
         image: elotl/virtual-kubelet:latest
         imagePullPolicy: Always
         name: virtual-kubelet
@@ -103,10 +104,29 @@ spec:
           readOnly: true
         securityContext:
           privileged: true
+      - command:
+        - /bin/sh
+        - -c
+        - exec kube-proxy --oom-score-adj=-998 --bind-address=127.0.0.1 --v=2
+        image: k8s.gcr.io/kube-proxy:v1.18.3
+        imagePullPolicy: IfNotPresent
+        name: kube-proxy
+        resources:
+          requests:
+            cpu: 100m
+        securityContext:
+          privileged: true
+        terminationMessagePath: /dev/termination-log
+        terminationMessagePolicy: File
+        volumeMounts:
+        - mountPath: /run/xtables.lock
+          name: xtables-lock
+        - mountPath: /lib/modules
+          name: lib-modules
+          readOnly: true
       dnsPolicy: ClusterFirst
       restartPolicy: Always
       serviceAccountName: virtual-kubelet
-      hostNetwork: true
       tolerations:
       - key: "node-role.kubernetes.io/master"
         effect: "NoSchedule"
@@ -123,7 +143,7 @@ spec:
             mode: 0600
       - name: xtables-lock
         hostPath:
-          path: /run/xtables.lock
+          path: /run/virtual-kubelet-xtables.lock
           type: FileOrCreate
       - name: lib-modules
         hostPath:
--- a/deploy/manifests/virtual-kubelet/base/sa.yaml
+++ b/deploy/manifests/virtual-kubelet/base/sa.yaml
@@ -105,6 +105,8 @@ rules:
       - endpoints
     verbs:
       - get
+      - list
+      - watch
   - apiGroups:
     - certificates.k8s.io
     resourceNames:
--- a/pkg/server/getstatssummary.go
+++ b/pkg/server/getstatssummary.go
@@ -29,10 +29,29 @@ import (
 	stats "k8s.io/kubernetes/pkg/kubelet/apis/stats/v1alpha1"
 )
 
+const (
+	nanosecondsPerSecond = float64(time.Second / time.Nanosecond)
+)
+
 type usageMetrics struct {
-	UsageNanoCores  uint64
-	UsageBytes      uint64
-	WorkingSetBytes uint64
+	UsageNanoCores       uint64
+	UsageCoreNanoSeconds uint64
+	AvailableBytes       uint64
+	UsageBytes           uint64
+	WorkingSetBytes      uint64
+	RSSBytes             uint64
+	PageFaults           uint64
+	MajorPageFaults      uint64
+	NetRxBytes           uint64
+	NetRxErrors          uint64
+	NetTxBytes           uint64
+	NetTxErrors          uint64
+	FSAvailableBytes     uint64
+	FSCapacityBytes      uint64
+	FSUsedBytes          uint64
+	FSInodesFree         uint64
+	FSInodes             uint64
+	FSInodesUsed         uint64
 }
 
 func (p *InstanceProvider) GetStatsSummary(ctx context.Context) (*stats.Summary, error) {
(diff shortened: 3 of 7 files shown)


justnoise:
One thing to keep in mind is that all these values are kept in memory and we keep 240 datapoints per pod.  I think we're still pretty good with the numbers.  A rough calculation:

- 240 datapoints/pod
- 400 pods
- conservatively 30 datapoints per pod stored in a map[string]float64

A quick browse shows that a map has an overhead in the ballpark of 30-60 bytes per entry + storage for the string key and int64 values.

All of that means that when running at scale, our stats should require 100MB - 500MB.  The larger number assumes a bunch of sidecars in each pod.  I think that's fine.  I sure hope that calculation was right!

deploy/manifests/virtual-kubelet/base/deployment.yaml
```diff
@@ -103,10 +104,29 @@ spec:
           readOnly: true
         securityContext:
           privileged: true
+      - command:
```
justnoise, line 107: I do want to try attaching AWS and GCE cloud load balancers to services to make sure this doesn't break anything.  It _should_ work but I've said that about code a lot and been wrong a lot.

deploy/manifests/virtual-kubelet/base/deployment.yaml
```diff
@@ -103,10 +104,29 @@ spec:
...
+        volumeMounts:
+        - mountPath: /run/xtables.lock
+          name: xtables-lock
+        - mountPath: /lib/modules
+          name: lib-modules
+          readOnly: true
       dnsPolicy: ClusterFirst
       restartPolicy: Always
       serviceAccountName: virtual-kubelet
-      hostNetwork: true
```
justnoise, line 109: I just realized this might be an issue?  Maybe?  How does this change interact with being able to reach the metadata service for AWS and GCE?

ldx: That still works:

GCE:

    $ kubectl -n kube-system exec -ti virtual-kubelet-84584bd8fc-4zx8q -c virtual-kubelet -- bash
    bash-5.0# curl -H "Metadata-Flavor: Google" http://169.254.169.254/computeMetadata/v1/instance/name; echo
    gke-vilmos-node-pool-vilmos-2ae4f1eb-zfj1

AWS:

    $ kubectl -n kube-system exec -ti virtual-kubelet-776f968878-hbddt -c virtual-kubelet -- bash
    bash-5.0# curl http://169.254.169.254/latest/meta-data/instance-id; echo
    i-017c0ab709a2201ee

deploy/manifests/virtual-kubelet/base/deployment.yaml
ldx: > I do want to try attaching AWS and GCE cloud load balancers to services to make sure this doesn't break anything. It _should_ work but I've said that about code a lot and been wrong a lot.

That works on AWS:

    $ kubectl run nginx --image=nginx
    pod/nginx created
    $ kubectl expose pod nginx --port=80 --type=LoadBalancer
    service/nginx exposed
    $ kubectl get svc nginx
    NAME    TYPE           CLUSTER-IP       EXTERNAL-IP                                                               PORT(S)        AGE
    nginx   LoadBalancer   10.105.117.135   a64d230df08874c8399a98e115f0e3a2-1918165344.us-west-1.elb.amazonaws.com   80:32168/TCP   43m
    $ curl -fs a64d230df08874c8399a98e115f0e3a2-1918165344.us-west-1.elb.amazonaws.com | grep title
    <title>Welcome to nginx!</title>

However, the GCE cloud provider in k8s does not like the fact that there's no cloud instance called virtual-kubelet:

    default     3s          Normal    EnsuringLoadBalancer         service/nginx          Ensuring load balancer
    default     3s          Warning   CreatingLoadBalancerFailed   service/nginx          Error creating load balancer (will retry): failed to ensure load balancer for service default/nginx: instance not found

Whether we use host network mode or not does not matter though. It's hard-coded in the GCE cloud plugin to list all nodes and bail if any of them is not a GCE VM instance: https://github.com/kubernetes/legacy-cloud-providers/blob/master/gce/gce_loadbalancer_external.go#L59-L62

I'll create an issue for it.

ldx:
> One thing to keep in mind is that all these values are kept in memory and we keep 240 datapoints per pod. I think we're still pretty good with the numbers. A rough calculation:
> 
>     * 240 datapoints/pod
> 
>     * 400 pods
> 
>     * conservatively 30 datapoints per pod stored in a map[string]float64
> 
> 
> A quick browse shows that a map has an overhead in the ballpark of 30-60 bytes per entry + storage for the string key and int64 values.
> 
> All of that means that when running at scale, our stats should require 100MB - 500MB. The larger number assumes a bunch of sidecars in each pod. I think that's fine. I sure hope that calculation was right!

Thanks! I think, with certain workloads, it might even be slightly more memory, since we also have per-container metrics. Something to keep in mind when scaling or provisioning a production cluster.

justnoise:
Thanks so much for testing metadata, LBs out and finding the issue with GCE!  I realize it was likely a bit time consuming.  I hope that moving away from hostNetwork mode will make things a bit smoother overall.

PR #108, opened 2020-05-28
Bcox gke metadata

justnoise:
This adds the following GKE instance metadata keys present on the virtual-kubelet node to the instance metadata of Kip Cells:

- cluster-location
- cluster-name
- cluster-uid

After these were added, DaemonSet Pods now start up and run but with errors about service accounts in the logs.

I also cleaned up initialization of our gceClient and did a minor commit to change the order layers are created in our Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,9 @@ FROM alpine
 
 RUN apk add --update bash ca-certificates iptables
 
-COPY virtual-kubelet /virtual-kubelet
-RUN chmod 755 /virtual-kubelet
 COPY kipctl /kipctl
 RUN chmod 755 /kipctl
+COPY virtual-kubelet /virtual-kubelet
+RUN chmod 755 /virtual-kubelet
 
 ENTRYPOINT ["/virtual-kubelet"]
--- a/pkg/server/cloud/gce/metadata.go
+++ b/pkg/server/cloud/gce/metadata.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"cloud.google.com/go/compute/metadata"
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -2,9 +2,9 @@ FROM alpine
 
 RUN apk add --update bash ca-certificates iptables
 
-COPY virtual-kubelet /virtual-kubelet
-RUN chmod 755 /virtual-kubelet
 COPY kipctl /kipctl
 RUN chmod 755 /kipctl
+COPY virtual-kubelet /virtual-kubelet
+RUN chmod 755 /virtual-kubelet
```
ldx, line 8: Great idea :ok_hand:

pkg/server/cloud/gce/metadata.go
```diff
@@ -80,3 +81,31 @@ func getZoneFromMetadata(c *metadata.Client) (string, error) {
...
+			// might not run on our virtual node but we are running
+			// them mostly for apprarances sake.  They should be
+			// patched to not run on our kip nodes.
+			klog.Warningln("unable to retrieve gke metadata key", k)
+			continue
+		}
+		gkeMD[k] = val
+	}
+	return gkeMD
+}
```
ldx, line 111: Do we handle the case when Kip is running outside of GCE (minikube or on-prem)? I guess it'll just log that the metadata service is not available and move on?

justnoise: I don't think we should take care of that case at this time.  If we're not inside GCE, we'll assume we're not on GKE.  If that assumption doesn't hold up we can make changes.

PR #111, opened 2020-05-29
Enable Kip Stateful Sets

### LINKED ISSUE #52: Example manifests need to protect kip's internal storage/state
Kip keeps an internal storage of:

- Identity
- CA
- Pods and Nodes

If a kip deployment is restarted without preserving the PersistentVolume kip's state is stored on, all pods will be restarted and the previous kip cells will be orphaned in the cloud since kip's identity will be recreated as well.  We've talked about this in other discussions in slack.

Todo:

- [x] Ensure our deployment scripts recreate the persistent volume that stores kip's data
- [x] Explore using a statefulSet as a method of ensuring virtual-kubelet identity. That would require creating a cluster identity in the API server when starting up kip and then using that plus the stateful identity to identify cloud resources that belong to an instance of a kip instance.

jrroman:
Closes #52 

- Added project as a field in the env.tfvars.example so users dont have to type it everytime they run terraform apply addressed in commit 47d3aba
- Added `.terraform` directory to .gitignore for the `/deploy/terraform-gcp/` directory in commit 4f8ebfd 
- Have updated the VK kip deployments to become stateful in commit 55f98a2


deploy/manifests/virtual-kubelet/base/deployment.yaml
```diff
@@ -1,17 +1,30 @@
...
+---
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
   name: virtual-kubelet
   namespace: kube-system
+  labels:
+    app: virtual-kubelet
 spec:
+  serviceName: virtual-kubelet
```
ldx, line 23: Shouldn't this reference the service created above? It is called vk-service in the service manifest.

deploy/manifests/virtual-kubelet/base/deployment.yaml
```diff
@@ -1,17 +1,30 @@
...
   labels:
-    app: virtual-kubelet
+    app: vk-service
+spec:
+  ports:
+  - port: 80
+    name: vk-headless-service
+  clusterIP: None
+  selector:
+    app: vk-service
```
ldx, line 13: I think this should reference the app label in the StatefulSet template, shouldn't it? It's app=virtual-kubelet there.

jrroman: Yes you are correct in both of these cases i have updated the service to just use virtual-kubelet as the name.

deploy/manifests/virtual-kubelet/base/statefulset.yaml
```diff
@@ -1,17 +1,30 @@
-apiVersion: apps/v1
-kind: Deployment
+apiVersion: v1
```
justnoise, line 1: What's the purpose of the virtual-kubelet Service?

deploy/manifests/virtual-kubelet/base/statefulset.yaml
```diff
@@ -149,3 +162,12 @@ spec:
...
           type: ""
+  volumeClaimTemplates:
+    - metadata:
+        name: data
+      spec:
+        accessModes: 
+          - ReadWriteOnce
+        resources:
+          requests:
+            storage: 1Gi
```
justnoise, line 173: Can we increase the size of this?  Typically our etcd data is ~ 500MB, even for small deployments.  We might want to go anywhere from 3-5GB, an etcd volume can grow up to 10GB.

deploy/manifests/virtual-kubelet/base/statefulset.yaml
jrroman: So when using a statefulset `serviceName` is a required field for it, so we need a headless service to be included as well.

According to Kubernetes documentation "StatefulSets currently require a Headless Service to be responsible for the network identity of the Pods. You are responsible for creating this Service."

The above can be found in this document in the "Limitations" section.
https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/

Issue #92, opened 2020-05-20
Associate GCE service account with logging and monitoring API access to cell VM instances


ldx:
Applications running on the VM use the service account of the instance to call Google Cloud APIs. We need a service account for cells that has the permissions "logs.write" and "monitoring.write", so that we can run the monitoring and logging agents.

ldx:
https://cloud.google.com/monitoring/agent/authorization
https://cloud.google.com/logging/docs/agent/authorization

ldx:
Fixed in #114

PR #114, opened 2020-06-01
Add logging.write and monitoring.write scopes to GCE instances

ldx:
We plan to run a monitoring and a logging agent on GCE instances to
better integrate with stackdriver. This adds the necessary scopes so
that those agents can send data to SD.


ldx:
This fixes #92

PR #115, opened 2020-06-01
Build multiple init-cert images

ldx:
Since init-cert uses kubectl, we need to ensure the right version is
used (kubectl guarantees one minor version skew compatibility).

I added a script to build multiple images for each Kubernetes version
between 1.12 and 1.18. Depending on the cluster, the use can simply
change the image to e.g. elotl/init-cert:k8s-1.14 in a 1.14.x cluster.

### DIFF
--- a/scripts/init-cert/Dockerfile
+++ b/scripts/init-cert/Dockerfile
@@ -1,5 +1,10 @@
 FROM debian:stable-slim
 
+ARG K8S_VERSION_MAJOR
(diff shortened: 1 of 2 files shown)


ldx:
> I think this is alright. I assume we'll have to manually specify the k8s version in the virtual-kubelet deployment manifest?

It's optional, by default the latest available kubectl version will be used (i.e. the latest tag will point to the image with the latest kubectl version).

PR #122, opened 2020-06-05
Update README.md

ldx:
- Updated README with new naming.
- Fixed pod anti-affinity for the statefulset.
- Renamed terraform to terraform-aws.


README.md
```diff
@@ -16,61 +16,106 @@ When workloads run on Kip, your cluster size naturally scales with the cluster w
...
+There are two ways to get Kip up and running.
+
+* Option 1: Use the provided terraform scripts to create a new Kubernetes cluster with a single Kip node on AWS or GKE.
 * Option 2: Add Kip to an existing kubernetes cluster.
 
 ### Installation Option 1: Create a Minimal K8s Cluster
 
 Prequisites:
-- An AWS account
+- An AWS or GCE account
```
jrroman, line 28: Naming with google is confusing do you think we should call this GCP since the "account" is for all of google cloud. Just thinking from an end user perspective.

ldx: Even Google is inconsistent with hits... For example, their ingress is called ingress-gce, while their k8s cloud provider is cloud-provider-gcp. Probably GCP is better as you pointed out, let me change it.

PR #130, opened 2020-06-15
Create init-cert image in build, and update latest tags after tests

ldx:
Changes:
- Build only one init-cert image, with all supported kubectl versions included. One that is compatible with the server is chosen during runtime.
- Have the build create and use an init-cert image, instead of pulling an existing one from the registry.
- If all tests pass, and the build is for a release tag in git, update the :latest tag for elotl/kip and elotl/init-cert.

### DIFF
--- a/scripts/init-cert/build.sh
+++ b/scripts/init-cert/build.sh
@@ -11,7 +11,11 @@ UPDATE_LATEST=${UPDATE_LATEST:-false}
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd $SCRIPT_DIR
 
-docker build -t ${REPO}:${TAG} --build-arg K8S_VERSIONS="1.14 1.15 1.16 1.17 1.18" .
+LATEST_K8S_VERSION=$(curl -fsL https://storage.googleapis.com/kubernetes-release/release/stable.txt)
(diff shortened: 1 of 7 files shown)


scripts/init-cert/build.sh
```diff
@@ -2,14 +2,21 @@
...
-docker tag elotl/init-cert:k8s-1.18 elotl/init-cert:latest
-$PUSH_IMAGES && docker push elotl/init-cert:latest || true
+REPO=${REPO:-elotl/init-cert}
+PUSH_IMAGES=${PUSH_IMAGES:-true}
+UPDATE_LATEST=${UPDATE_LATEST:-false}
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd $SCRIPT_DIR
+
+docker build -t ${REPO}:${TAG} --build-arg K8S_VERSIONS="1.14 1.15 1.16 1.17 1.18" .
```
justnoise, line 14: Since updating versions is always a pain and we are already automating the heck out of things, K8S_VERSIONS could also be auto-generated:
```
latest=$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)
latest_minor=$(echo $latest | sed -r 's/v1\.([0-9]+).*/\1/')
k8s_versions=$(seq -f '1.%g' -s ' ' 14 $latest_minor)
echo k8s_versions
```
Not sure if you'd like to add that complexity or whether updating the version string is something that might be a pain to keep track of.  I just got a little jealous of `kubectl.sh` below.

ldx: Good idea :+1:

PR #138, opened 2020-06-26
updating docs and manifests for deploying into GCE

justnoise:
Kip can use the service account credentials on its host instance. Updating manifests and docs to reflect this.

1. If deploying via terraform, change the default kustomize base to GCE and default to pulling credentials from the host instance.
2. If adding kip to an existing k8s cluster, update the docs to talk about 2 ways of authenticating (similar to AWS)

Also updated docs/provider-yaml.md to have a blurb about GCE cloud configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,9 +62,9 @@ You can configure the AWS access key Kip will use in your provider configuration
 
 In AWS, Kip can use credentials supplied by the [instance profile](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html) attached to the node the pod is dispatched to.  To use an instance profile, create an IAM policy with the [minimum Kip permissions](docs/kip-iam-permissions.md) then apply the instance profile to the node that will run the Kip provider pod.  The Kip pod must run on the cloud instance that the instance profile is attached to.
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -62,9 +62,13 @@ You can configure the AWS access key Kip will use in your provider configuration
 
 In AWS, Kip can use credentials supplied by the [instance profile](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html) attached to the node the pod is dispatched to.  To use an instance profile, create an IAM policy with the [minimum Kip permissions](docs/kip-iam-permissions.md) then apply the instance profile to the node that will run the Kip provider pod.  The Kip pod must run on the cloud instance that the instance profile is attached to.
 
+**GCP Credentials Option 1 - Built-in instance service account:**
+
+In GCE, Kip can use the service account built into the instance.  Kip requires https://www.googleapis.com/auth/compute scope in order to launch instances.
```
ldx, line 67: Wondering if "built into the instance" is the right way to put it? May "scope associated to the instance"?

justnoise: The docs call it the "built-in service account" but I think I took that and used it the wrong way.  I'll just say "the service account attached to the instance".

PR #140, opened 2020-06-30
Remove script pushing to GCR and ECR image repos

ldx:
Edit: changed this to remove script pushing to GCR and ECR image repos.


ldx:
Thinking more about this, since we have images available via the Marketplace and also staging image repositories, this is redundant, I'll remove this script.

Issue #141, opened 2020-07-01
I have a private AWS setup an I want to change the AWS URL which kip is trying to connect


saiharshitachava:
I have a private AWS setup an I want to change the AWS URL..How do I do that

Right now I guess its going to public AWS env..I see options of key ID and  region can I explicitly set URL as well? 

Error: error initializing provider kip: error configuring cloud client: Error setting up cloud client: Could not configure AWS cloud client authorization: Error validationg connection to AWS: RequestError: send request failed
caused by: Post https://ec2.us-east-1.amazonaws.com/: x509: certificate signed by unknown authority

@myechuri

saiharshitachava:
I did try giving the VPC ID in my provider yaml..dint work out..
Any help here @myechuri  @justnoise

myechuri:
@saiharshitachava : can you please clarify the following?

1) How did you deploy kip? 

a) Did you provision a new cluster using https://github.com/elotl/kip/tree/master/deploy/terraform-aws by following documentation at https://github.com/elotl/kip#installation-option-1-create-a-minimal-k8s-cluster? 

Or 

b) Did you deploy kip onto an existing cluster using https://github.com/elotl/kip#installation-option-2-using-an-existing-cluster ? If you used this option, did you use `AWS Credentials Option 1 - Configuring AWS API keys` or `AWS Credentials Option 2 - Instance Profile Credentials`?

2) Can you please share contents of your `provider.yaml` after redacting sensitive information? Thanks!

saiharshitachava:
@myechuri I deployed kip on existing cluster using the files in base
I used the option of api keys

apiVersion: v1
cloud:
  aws:
    region: "us-east-1"
    accessKeyID: "******"
    secretAccessKey: "******"
    vpcID: "Gave my internal VPC ID found in AWS VPC section"
etcd:
  internal:
    dataDir: /opt/kip/data
cells:
  bootImageSpec:
    owners: 68949425850dd1
    filters: name=elotl-kip-*
  defaultInstanceType: "t3.small"
  defaultVolumeSize: "2G"
  nametag: kip
  itzo:
    url: https://itzo-kip-download.s3.amazonaws.com
    version: latest
kubelet:
  cpu: "2"
  memory: "2Gi"
  pods: "2"

myechuri:
Thanks, @saiharshitachava . Can you please clarify if you filled in `accessKeyID` and `secretAccessKey` in your provider.yaml or did you leave them empty?

```
[...]

PR #143, opened 2020-07-01
allow user to specify endpointURL in provider.yaml. 

justnoise:
Addresses the point in issue in https://github.com/elotl/kip/issues/141.  Allow the user to specify a custom endpoint URL used by the ec2 client.  This overrides the default behavior of connecting to the endpoint at `protocol://service-code.region-code.amazonaws.com` (see: https://docs.aws.amazon.com/general/latest/gr/rande.html).

Also get rid of the unused AWS ELB Client (from the old Milpa days).

### DIFF
--- a/docs/provider-config.md
+++ b/docs/provider-config.md
@@ -41,6 +41,10 @@ cloud:
     # and will launch pods into that subnet.
     # subnetID: ''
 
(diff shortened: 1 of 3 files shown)


saiharshitachava:
Do we need to build a new image of kip or something for us to use endpointurl?

@justnoise Thanks for merging it

PR #147, opened 2020-07-03
Bcox aws insecure tls skip verify

justnoise:


pkg/server/cloud/aws/aws.go
```diff
@@ -58,36 +58,48 @@ type AwsEC2 struct {
...
-		return nil, util.WrapError(err, "Error creating EC2 client session")
+func getAWSConfig(endpointURL string, insecureSkipSSLVerify bool) *aws.Config {
+	httpClient := &http.Client{
+		Timeout: awsTimeout,
 	}
 
+	// Disabling security checks is dangerous and should be
+	// avoided. Use case for this one was an on-prem proxy with self
+	// signed certificate.  We might consider only enabling this option
+	// if enpointURL is also not empty
```
myechuri, line 69: Is there a reason to not check for endpointURL not being empty right now? Thanks!

justnoise: I think the reasoning is that we're adding this feature without entirely knowing how it will be used.  An on-premise network could route `us-east-1.amazonaws.com` traffic through a local proxy, etc.  We just don't know how this will be used.

PR #157, opened 2020-07-17
Use kubeconfig in cell controller

ldx:
Allow using a kubeconfig file for API server access. The main change is to configure the cell controller via kubeconfig instead of using InClusterConfig() if a kubeconfig file is provided to Kip.

I also refactored the way we create the network agent kubeconfig, adding support for loading the same or another kubeconfig file instead of creating one from a service account secret.

Node-cli only supports an environment variable and no command line option:

        env:
        - name: KUBECONFIG
          value: /etc/kip/kubeconfig

For the network agent kubeconfig, I added `--network-agent-kubeconfig`.

Example, using one kubeconfig file that will be used by Kip and also the network agent on cells:

      containers:
      - command:
        - /kip
        - --provider
        - kip
        - --provider-config
        - /etc/kip/provider.yaml
        - --network-agent-kubeconfig
        - /etc/kip/kubeconfig
        env:
        - name: KUBECONFIG
          value: /etc/kip/kubeconfig


cmd/kip/kubeconfig.go
```diff
@@ -0,0 +1,42 @@
...
+	return kubeConfig, nil
+}
+
+func loadOrCreateKubeConfig(path, mainKubeConfig, saSecret string, rm *manager.ResourceManager) (*clientcmdapi.Config, error) {
+	if path == "" && saSecret == "" {
+		return nil, fmt.Errorf("no kubeconfig or service account secret provided")
+	}
+	if path != "" {
+		if saSecret != "" {
+			return nil, fmt.Errorf("no kubeconfig or service account secret provided")
```
myechuri, line 29: Should this message be different since we are here in case of path not being null and service account secret not being null? Thanks.

ldx: Correct, this was a copy/paste problem. Thanks!

PR #158, opened 2020-07-27
Bcox ecr multi region

justnoise:
ECR Registry credentials are issued regionally.  Update Kip to create creds in the region the image is stored in.

### DIFF
--- a/pkg/server/cloud/functional/functional_test_template.go
+++ b/pkg/server/cloud/functional/functional_test_template.go
@@ -173,8 +173,14 @@ func ContainerAuthTest(t *testing.T, c cloud.CloudClient) {
 	assert.NoError(t, err, "Error getting container authorization")
 	assert.Equal(t, "AWS", username1)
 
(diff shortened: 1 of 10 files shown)


pkg/server/cloud/functional/functional_test_template.go
```diff
@@ -169,12 +169,18 @@ func deleteInstances(t *testing.T, c cloud.CloudClient) {
 }
 
 func ContainerAuthTest(t *testing.T, c cloud.CloudClient) {
-	username1, password1, err := c.GetRegistryAuth()
+	username1, password1, err := c.GetRegistryAuth("689494258501.dkr.ecr.us-east-1.amazonaws.com/helloserver:latest")
```
myechuri, line 172: Is this test expected to run on GCP? Thanks.

justnoise: Good point, while this is part of the CloudClient interface, this test is very AWS specific.  It should be moved into the `aws` package.  We currently don't have support for automatically creating secrets in GCP since it appears to be more complex than just generating a token.

PR #167, opened 2020-08-20
Add docs on secondary IPs

ldx:


ldx:
> Thanks @ldx . Can we please add the suggested blob you have for SLES as an example? Thanks!

I would rather not. Documenting the behavior of a specific operating system version is out of scope for us, not something we have tested or plan to test, and thus makes it more of a liability than an asset. The documentation points out that if one wants to use their custom environment, what configuration they need to watch out for.

PR #170, opened 2020-08-27
capture root disk size

hidalgopl:


pkg/server/cloud/common.go
```diff
@@ -140,3 +140,10 @@ func ToSaneVolumeSize(volSizeSpec string) int32 {
 	}
 	return volSizeGiB
 }
+
+func AddVolSpecSizeToRootSize(specSize int32, img Image) int32 {
+	if img.VolumeDiskSize == nil {
+		return specSize
+	}
+	return *img.VolumeDiskSize + specSize
```
justnoise, line 148: Peeking in on your work in progress: I think what we want to do here is make sure we don't request a disk size (the `specSize`) that's less than the actual volume size for the disk since that causes boot errors. No need to add, just return the larger of `img.VolumeDiskSize` and `specSize`

pkg/server/cloud/aws/instances.go
```diff
@@ -231,6 +232,25 @@ func bootImageSpecToDescribeImagesInput(spec cloud.BootImageSpec) *ec2.DescribeI
 	return input
 }
 
+func getRootDeviceName(img *ec2.Image) (string, error) {
+	if img.RootDeviceName == nil {
+		msg := "cannot get root device name from image: " + *img.Name
+		return "", errors.New(msg)
+	}
+	return *img.RootDeviceName, nil
```
ldx, line 240: The preferred way is to use `aws.String()` and `aws.StringValue()` for dealing with pointers in the data structures in the AWS SDK.

pkg/server/cloud/aws/instances.go
```diff
@@ -231,6 +232,25 @@ func bootImageSpecToDescribeImagesInput(spec cloud.BootImageSpec) *ec2.DescribeI
...
+}
+
+func getRootDeviceVolumeSize(blockDevices []*ec2.BlockDeviceMapping, rootDeviceName string) *int32 {
+	var rootDiskSize int32
+	for _, blockDevice := range blockDevices {
+		if *blockDevice.DeviceName == rootDeviceName {
+			rootDiskSize = int32(*blockDevice.Ebs.VolumeSize)
```
ldx, line 247: Not sure if these pointers are ever `nil`, but I'd also use the `aws.*Value()` helpers here.

pkg/server/cloud/aws/instances.go
```diff
@@ -255,11 +275,19 @@ func (e *AwsEC2) GetImage(spec cloud.BootImageSpec) (cloud.Image, error) {
 				creationTime = &ts
 			}
 		}
+		rootDeviceName, err := getRootDeviceName(img)
+		if err != nil {
+			klog.Warningf("getting root device name from image: %v", err.Error())
+		}
+		var rootDiskSize *int32
+		rootDiskSize = getRootDeviceVolumeSize(img.BlockDeviceMappings, rootDeviceName)
```
ldx, line 283: Nitpicking: you don't need to declare the variable separately if you're assigning a value to it right away.

pkg/server/cloud/aws/instances_test.go
```diff
@@ -67,3 +67,51 @@ func TestBootImageSpecToDescribeImagesInput(t *testing.T) {
...
+						DeleteOnTermination: nil,
+						Encrypted:           nil,
+						Iops:                nil,
+						KmsKeyId:            nil,
+						SnapshotId:          nil,
+						VolumeSize:          &volumeSize,
+						VolumeType:          nil,
+					},
+					NoDevice:    nil,
+					VirtualName: nil,
```
ldx, line 103: You can help readability here if leave out the pointers that are `nil`, since that's their default value.

pkg/server/cloud/aws/instances_test.go
```diff
@@ -67,3 +67,51 @@ func TestBootImageSpecToDescribeImagesInput(t *testing.T) {
...
+				},
+			},
+			"root-device",
+			&expectedVolumeSize,
+
+		},
```
ldx, line 109: Can you also add test cases where the volume is not found, and where the list of volumes is empty?

pkg/server/cloud/gce/instances.go
```diff
@@ -528,12 +528,14 @@ func (c *gceClient) GetImage(spec cloud.BootImageSpec) (cloud.Image, error) {
...
 		}
 	}
+	diskSize := int32(resp.DiskSizeGb)
 	// TODO: these values seem to be reversed?
 	return cloud.Image{
 		ID:           resp.Name,
 		Name:         resp.SelfLink,
 		RootDevice:   "",
 		CreationTime: creationTime,
+		VolumeDiskSize: &diskSize,
```
ldx, line 538: We should make sure to test this on an actual GKE cluster.

pkg/server/cloud/aws/instances_test.go
hidalgopl: Good point, added :+1:

pkg/server/cloud/aws/instances.go
hidalgopl: Thanks :+1:  switched to using those.

pkg/server/cloud/aws/instances.go
hidalgopl: nice catch:)

pkg/server/cloud/aws/instances_test.go
hidalgopl: removed those nils ;)

pkg/server/cloud/gce/instances.go
hidalgopl: I'm happy to do this, I'll probably need a bit of helping in setting up the cluster

pkg/server/cloud/aws/instances.go
```diff
@@ -231,6 +231,18 @@ func bootImageSpecToDescribeImagesInput(spec cloud.BootImageSpec) *ec2.DescribeI
 	return input
 }
 
+func getRootDeviceVolumeSize(blockDevices []*ec2.BlockDeviceMapping, rootDeviceName string) int32 {
+	var rootDiskSize int32
+	for _, blockDevice := range blockDevices {
+		if *blockDevice.DeviceName == rootDeviceName {
```
justnoise, line 237: When dealing with AWS data structures and values, I've found it safer to follow the assumption: "Almost everything can have bad data". If a cloud provider pushes out a change to their API that has a bug (rare but can happen) and that bug returns a nil value for `DeviceName` or `blockDevice.Ebs`, Kip will crash. Also there can be times where something you thought "This can't ever be nil" turns out to be nil. E.g: https://github.com/elotl/kip/pull/173

It kinda stinks but it's better to be defensive against these types of errors.  I would suggest the following change to protect against nil values:

```go
		if aws.StringValue(blockDevice.DeviceName) == rootDeviceName &&
			blockDevice.Ebs != nil {
			rootDiskSize = aws.Int32Value(blockDevice.Ebs.VolumeSize)
			break
		}
```

pkg/server/cloud/common.go
```diff
@@ -138,5 +138,10 @@ func ToSaneVolumeSize(volSizeSpec string) int32 {
 		klog.Errorln("Empty volume size found in resource spec, setting to reasonable value")
 		volSizeGiB = 8
 	}
+	if image.VolumeDiskSize != nil {
```
justnoise, line 141: It's a minor issue: I would move this check ahead of the `if volSizeGiB == 0 {` check.  That check is a bit of a fail-safe to ensure users don't end up with a zero-sized volume. I also believe it still gets used in our functional tests.  The best  order is to first choose `*image.VolumeDiskSize` if it is available and, if not, to fall back to the arbitrarily chosen size of 8GB.  In other words: Do the thing that makes the most sense first and, if that fails, do something we only hope works.

pkg/server/cloud/cloud.go
```diff
@@ -109,6 +109,7 @@ type Image struct {
 	Name         string
 	RootDevice   string
 	CreationTime *time.Time
+	VolumeDiskSize *int32 // in GiB
```
justnoise, line 112: Random idea, doesn't have to be done.  This could also be an `int32` instead of an `*int32`. It would default to 0 and, since volumes can't have a size of zero, the zero sized voulume would be equivalent to a nil value if you were to use a pointer.

The only real advantage of this is that you don't have to test for a non-nil value before using the value in the comparison in `if *image.VolumeDiskSize > volSizeGiB`.  Also, you don't need to dereference the value.

justnoise (changes requested):
Nice work.  There are a couple of changes needed.  When reviewing other PRs we would usually point out the necessary changes but, if the logic and implementation are otherwise good, we approve the PR.

I'm requesting changes here just to work through everything.  This is your first PR so I'm a bit more conservative.

pkg/server/cloud/gce/instances.go
ldx: > I'm happy to do this, I'll probably need a bit of helping in setting up the cluster

Sure, let's talk about it once this PR wraps up.

pkg/server/cloud/aws/instances.go
hidalgopl: Makes sense, I've added check if Ebs is nil. I kept `rootDiskSize = int32(blockDevice.Ebs.VolumeSize)` as originally `blockDevice.Ebs.VolumeSize` is int64, so we have to convert it.

pkg/server/cloud/common.go
hidalgopl: Good point, moved this if block :)

pkg/server/cloud/cloud.go
hidalgopl: I originally used pointer, as I thought it's pretty standard way to deal with optional values in struct. However, defaulting to 0 makes total sense to me. Thanks!

pkg/server/cloud/aws/instances.go
justnoise: In that case the same point stands but with different data types, I would suggest `rootDiskSize = int32(aws.Int64Value(blockDevice.Ebs.VolumeSize))`.

What we've really been strongly recommending: it's good coding practice to use AWS's unpacking functions when dealing with their datastructures.  Directly dereferencing data from those structures without testing for nil comes with risks.

pkg/server/cloud/aws/instances.go
justnoise: To be a bit clearer: The main issue here is that `*blockDevice.Ebs.VolumeSize` dereferences `VolumeSize`. If that value is `nil`, this will panic.  Below there's a dance to take the address of `rootDiskSize` and pass that through `aws.Int32Value`.  That's not needed as `rootDiskSize` is already an int32 so that's basically a strange noop with the overhead of a function call.

justnoise (approved):
LGTM

pkg/server/cloud/gce/instances.go
hidalgopl: as we discussed on slack, I tested on GCE and it looks good, resolving :+1:

pkg/server/cloud/aws/instances.go
```diff
@@ -231,6 +231,17 @@ func bootImageSpecToDescribeImagesInput(spec cloud.BootImageSpec) *ec2.DescribeI
 	return input
 }
 
+func getRootDeviceVolumeSize(blockDevices []*ec2.BlockDeviceMapping, rootDeviceName string) int32 {
+	var rootDiskSize int32
+	for _, blockDevice := range blockDevices {
+		if *blockDevice.DeviceName == rootDeviceName && blockDevice.Ebs != nil {
```
justnoise, line 237: I know that I suggested this change but it also makes the mistake of dereferencing (the possibly nil) `blockDevice.DeviceName`.  To keep everything consistent, lets use `aws.StringValue` here as well.

hidalgopl: :+1: fixed

ldx (approved):
:+1:

PR #179, opened 2020-09-25
wait for volume modification status when resizing volume

justnoise:
Title says it all. We've been having some strange issues where resizing a volume fails in itzo because inside itzo the volume hasn't actually changed size.

This updates kip to wait for the volume modification state to change to `Optimizing` or `Complete`.

More info: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/monitoring-volume-modifications.html


ldx (approved):
:+1:

PR #180, opened 2020-09-30
Annotation addition for itzo to specify image overlay for rootfs

### LINKED ISSUE #178: Pass pod.elotl.co/ annotations through to itzo
We want to allow annotations to change how itzo does things.  Specifically we are interested in implementing this issue: https://github.com/elotl/itzo/issues/101

Here are some rough notes on how this can happen:
1. Annotations are already copied from k8s pods to kip pods in `pkg/server/convert.go:k8sToMilpaPod`. A suggested name for a new annotation was `image-overlay-rootfs` and the value should be a bool-ish value. All that needs to happen on the kip side is that annotation needs to get sent to itzo.
2. All pod creation/updates are sent to itzo in the `api.PodParameters` structure. Add an `Annotations map[string]string` to that structure
3. In the `PodController`, the pod spec and other info is sent to Itzo in `PodController.updatePodUnits()`.  Copy all `pod.elotl.co/*` pod annotations into the `PodParameters` that are sent to the itzo instance.
4. Now itzo gets all the pod annotations.

jrroman:
Closes #178 

**Updates:**
- Adding new annotation `pod.elotl.co/image-overlay-rootfs` 
- Adding a struct member to `api.PodParameters` called `Annotations` which is a `map[string]string`
- Created helper func to populate the Annotations map with all `pod.elotl.co/*` annotations. This map is then sent to itzo


pkg/api/annotations/pod.go
```diff
@@ -27,6 +27,14 @@ const PodLaunchType = "pod.elotl.co/launch-type"
 // override specified resource requests and limits.
 const PodInstanceType = "pod.elotl.co/instance-type"
 
+// PodImageOverlayRootfs is an annotation users add to their
+// kubernetes pods to inform kip whether the itzo cell will use
+// -mount or -extractto argument when itzo calls tosi to pull rootfs
+// images. This annotation will default to -mount if its value is empty
+// or truthy, if the value is falsy it will use the -extractto option.
+// Allowed values: 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False
+const PodImageOverlayRootfs = "pod.elotl.co/image-overlay-rootfs"
```
justnoise, line 36: @ldx: do we want to test this out and make sure it makes a difference before publishing docs on how to use it and making the annotation public, to last for all time?  If we do want this to be available to users, docs/annotations.md should be updated with the comment above.

pkg/api/annotations/pod.go
```diff
@@ -27,6 +27,14 @@ const PodLaunchType = "pod.elotl.co/launch-type"
 // override specified resource requests and limits.
 const PodInstanceType = "pod.elotl.co/instance-type"
 
+// PodImageOverlayRootfs is an annotation users add to their
+// kubernetes pods to inform kip whether the itzo cell will use
+// -mount or -extractto argument when itzo calls tosi to pull rootfs
+// images. This annotation will default to -mount if its value is empty
```
justnoise, line 33: Might want to add a sentence or modify the sentence to say what each of those arguments does and how the behavior of the system is different between the two.

justnoise (approved):

pkg/api/annotations/pod.go
ldx: I'd change this comment to something like:

    // PodImageOverlayRootfs is an annotation users add to their
    // kubernetes pods to inform kip whether to use overlayfs when
    // setting up the container images, or a directory with all the image layers
    // extracted. The default is overlayfs. Set the label value to a truthy value
    // for overlayfs, and to a falsy value to have image layers extracted for the
    // rootfs without overlayfs.

What parameters are used under the hood is not important for the user, and it's an implementation detail.

ldx (changes requested):

pkg/api/annotations/pod.go
jrroman: @ldx just so we can get this pr out of the way. Should I be adding this annotation to the `docs/annotations.md`? This PR is in reference to allowing the use of direct extraction of images vs using an overlayfs.

ldx (approved):

pkg/api/annotations/pod.go
ldx: Yeah, let's add it to annotations.md :+1:

PR #182, opened 2020-10-21
Attach AmazonSSMInstanceCore permissions to the master node

hidalgopl:


deploy/terraform-aws/main.tf
```diff
@@ -282,7 +282,32 @@ resource "aws_iam_role_policy" "k8s_node" {
...
+        "ssmmessages:CreateControlChannel",
+        "ssmmessages:CreateDataChannel",
+        "ssmmessages:OpenControlChannel",
+        "ssmmessages:OpenDataChannel",
+        "ec2messages:AcknowledgeMessage",
+        "ec2messages:DeleteMessage",
+        "ec2messages:FailMessage",
+        "ec2messages:GetEndpoint",
+        "ec2messages:GetMessages",
+        "ec2messages:SendReply"
```
ldx, line 310: This role policy is already pretty long since we've mixed permissions needed for Kip with permissions needed for the control plane + workers. How about unwinding it, and split it into 3 parts:
- Permissions for Kip (documented in docs/kip-iam-permissions.md),
- AWS permissions for the control plane + worker permissions from https://github.com/kubernetes/cloud-provider-aws
- SSM from the managed policy

Then we can attach these to the IAM role that we create.

ldx (changes requested):

deploy/terraform-aws/main.tf
```diff
@@ -179,8 +179,8 @@ EOF
   tags = local.k8s_cluster_tags
 }
 
-resource "aws_iam_role_policy" "k8s_node" {
-  name = "k8s-node-${var.cluster_name}"
+resource "aws_iam_role_policy" "k8s_node_k8s_policy" {
+  name = "k8s-node-${var.cluster_name}-k8s"
```
jrroman, line 183: nit: is the addition of `-k8s` after cluster name necessary since it is prefixed with the same?

jrroman (approved):
Was just curious about the one note I made below, I also like Vilmos's idea of splitting policies. 

Aside from that LGTM 👍

deploy/terraform-aws/main.tf
jrroman: Disregard this comment, since you have broken out the policy into three separate I now see the logic behind it.

deploy/terraform-aws/main.tf
```diff
@@ -293,6 +287,113 @@ resource "aws_iam_role_policy" "k8s_node" {
...
+EOF
+}
+
+
+resource "aws_iam_role_policy_attachment" "kip-policy-attachment" {
+  policy_arn = aws_iam_policy.k8s_node_kip_policy.arn
+  role = aws_iam_role.k8s_node.name
+}
+
+resource "aws_iam_policy" "k8s_node_ssm_policy" {
```
ldx, line 347: There's a managed IAM policy for SSM, we should use that one.

deploy/terraform-aws/main.tf
```diff
@@ -293,6 +287,113 @@ resource "aws_iam_role_policy" "k8s_node" {
...
+        "ec2:DeleteSecurityGroup",
+        "ec2:DescribeAvailabilityZones",
+        "ec2:DescribeDhcpOptions",
+        "ec2:DescribeImages",
+        "ec2:DescribeInstances",
+        "ec2:DescribeNetworkInterfaces",
+        "ec2:DescribeRouteTables",
+        "ec2:DescribeSecurityGroups",
+        "ec2:DescribeSubnets",
+        "ec2:DescribeVolumes",
```
ldx, line 319: Recently there was a change in Kip that now requires `ec2:DescribeVolumesModifications` too, can you add that both here and to docs/kip-iam-permissions.md? Thanks!

ldx (changes requested):
I dropped two more requests for changes, other than that it LGTM :+1:

deploy/terraform-aws/main.tf
hidalgopl: Added :+1:

deploy/terraform-aws/main.tf
hidalgopl: good point! changed it to `arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore` arn

ldx (approved):
Thanks :+1:

PR #188, opened 2020-11-13
Remove unused method from node client and itzo implementation for it

hidalgopl:
I guess it's kind of Milpa leftover (?)

### DIFF
--- a/pkg/nodeclient/itzoclient.go
+++ b/pkg/nodeclient/itzoclient.go
@@ -423,21 +423,6 @@ func (c *ItzoClient) Deploy(pod, name string, data io.Reader) error {
 	return nil
 }
 
-func (c *ItzoClient) RunCmd(cmdParams api.RunCmdParams) (string, error) {
-	url := c.baseURL + "rest/v1/runcmd/"
-	b, err := json.Marshal(cmdParams)
-	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/nodeclient/itzoclient.go
```diff
@@ -446,10 +431,6 @@ func StreamLogsEndpoint(unitName string, withMetadata bool) string {
 	return fmt.Sprintf("rest/v1/logs/%s?follow=1&metadata=%d", unitName, metadataflag)
 }
 
-func PortForwardEndpoint() string {
-	return fmt.Sprintf("rest/v1/portforward/")
-}
-
```
ldx, line 452: IIRC VK does not implement the port-forward endpoint, but we have implemented it in itzo. I'd keep it in here even though it's unused, it could a good and straightforward change in VK to add it to their API.

ldx (approved):
We don't use runcmd anymore, but port forwarding is something that we will need to thread through VK at some point, so I'd keep that one.

jrroman (approved):
:+1:

pkg/nodeclient/itzoclient.go
hidalgopl: :+1: reverted

PR #190, opened 2020-11-19
Ensure multiple ECR images can be pulled for a pod

ldx:
When iterating through images in the podspec, we need to make sure that:

1. We don't stop once the first ECR image is hit. In the podspec, there
   might be multiple images in different regions that need ECR
   credentials.
2. We take into account both regular containers and init containers.

### DIFF
--- a/pkg/server/pod_controller.go
+++ b/pkg/server/pod_controller.go
@@ -368,29 +368,33 @@ func (c *PodController) loadRegistryCredentials(pod *api.Pod) (map[string]api.Re
 
 	// AWS is different, they require us to authenticate with IAM
 	// Do that auth and pass along the username and password
-	for i := 0; i < len(pod.Spec.Units); i++ {
-		server, _, err := util.ParseImageSpec(pod.Spec.Units[i].Image)
+	if err := api.ForAllUnitsWithError(pod, func(unit *api.Unit) error {
+		image := unit.Image
+		server, _, err := util.ParseImageSpec(image)
 		if err != nil {
-			return nil, util.WrapError(err, "Could not parse image spec")
+			return util.WrapError(err, "Could not parse image spec")
 		}
 		if strings.HasSuffix(server, "amazonaws.com") {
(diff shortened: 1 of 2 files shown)


pkg/server/pod_controller.go
```diff
@@ -368,29 +368,33 @@ func (c *PodController) loadRegistryCredentials(pod *api.Pod) (map[string]api.Re
...
 			}
-			username, password, err := c.cloudClient.GetRegistryAuth(pod.Spec.Units[i].Image)
+			username, password, err := c.cloudClient.GetRegistryAuth(image)
 			if err != nil {
-				return nil, util.WrapError(err, "Could not get container auth")
+				return util.WrapError(err, "Could not get container auth")
 			}
 			allCreds[server] = api.RegistryCredentials{
 				Server:   string(server),
 				Username: string(username),
```
hidalgopl, line 390: nitpicking: I guess this conversion isn't needed, as cloudClient.GetRegistryAuth already returns strings

hidalgopl (approved):

jrroman (approved):
LGTM

pkg/server/pod_controller.go
```diff
@@ -368,29 +368,33 @@ func (c *PodController) loadRegistryCredentials(pod *api.Pod) (map[string]api.Re
 
 	// AWS is different, they require us to authenticate with IAM
 	// Do that auth and pass along the username and password
-	for i := 0; i < len(pod.Spec.Units); i++ {
-		server, _, err := util.ParseImageSpec(pod.Spec.Units[i].Image)
+	if err := api.ForAllUnitsWithError(pod, func(unit *api.Unit) error {
```
hidalgopl, line 371: we can consider in future giving this function a name and test it. Your call ;)

ldx: Good idea, I pulled it up into a separate function and added a test for it :+1:

PR #192, opened 2020-12-05
Use SSM Parameter Store on AWS to distribute instance parameters

ldx:
This is behind a config flag, so we can have users opt in; the default is still the old method of distributing the certificate and key to instances via instance user-data.

If enabled, the node controller will create an encrypted parameter with instance-specific configuration data such as the certificate, key, cell-config, etc. Itzo-launcher will retrieve it, deserialize the data and save the files, using the same paths as before.

I also refactored the cloudinitfile package to pull out the logic on specific files, so it's the same for cloud-init and SSM.

SSM access requires additional IAM permissions. I added them to the TF example, and also updated docs.

Example IAM policy in TF format for restricting access, so each instance can only read its own parameters:

    data "aws_iam_policy_document" "kip_cell_ssm_parameter_store" {
      version = "2012-10-17"
    
      statement {
        effect = "Allow"
        actions = [
          "ssm:GetParametersByPath",
        ]
        resources = [
          "arn:aws:ssm:us-west-2:689494258501:parameter/kip/cells/*",
        ]
      }
    
      statement {
        effect = "Allow"
        actions = [
          "ssm:GetParameter",
          "ssm:GetParameters",
        ]
        resources = [
          "arn:aws:ssm:us-west-2:689494258501:parameter/kip/cells/*",
        ]
        condition {
          test     = "StringEquals"
          variable = "aws:ResourceTag/AWSUserID"
          values = [
            "$${aws:userid}"
          ]
        }
      }
    
      statement {
        effect = "Deny"
        actions = [
          "ssm:GetParameter",
          "ssm:GetParameters",
        ]
        resources = [
          "arn:aws:ssm:us-west-2:689494258501:parameter/kip/cells/*",
        ]
        condition {
          test     = "StringNotEquals"
          variable = "aws:ResourceTag/AWSUserID"
          values = [
            "$${aws:userid}"
          ]
        }
      }
    }


pkg/server/config.go
```diff
@@ -138,20 +138,21 @@ type InternalEtcdConfig struct {
...
+	Itzo                   ItzoConfig                    `json:"itzo"`
+	ExtraCIDRs             []string                      `json:"extraCIDRs"`
+	ExtraSecurityGroups    []string                      `json:"extraSecurityGroups"`
+	Nametag                string                        `json:"nametag"`
+	StatusInterval         int                           `json:"statusInterval"`
+	HealthCheck            HealthCheckConfig             `json:"healthcheck"`
+	PrivateIPOnly          *bool                         `json:"privateIPOnly"`
+	CellConfig             map[string]string             `json:"cellConfig"`
+	DefaultIAMPermissions  string                        `json:"defaultIAMPermissions"`
+	UseCloudParameterStore bool                          `json:"useCloudParamaterStore"`
```
hidalgopl, line 155: could you this param with short explanation to provider docs as well?

pkg/server/nodemanager/node_controller.go
```diff
@@ -297,6 +285,33 @@ func (c *NodeController) startSingleNode(node *api.Node, image cloud.Image, clou
...
+
+	if c.Config.UseCloudParameterStore {
+		params, err := getMarshalledInstanceParameters(c.CertificateFactory, InstanceConfig{
+			ItzoURL:     c.Config.ItzoURL,
+			ItzoVersion: c.Config.ItzoVersion,
+			CellConfig:  c.Config.CellConfig,
+		})
+		if err != nil {
+			_ = c.stopSingleNode(node)
+			wrapErr := fmt.Errorf("createMarshalledNodeCertAndKey() for %s: %v", instanceID, err)
```
hidalgopl, line 298: Is this old name of `getMarshalledInstanceParameters` func? ;)

pkg/server/nodemanager/node_controller.go
```diff
@@ -183,34 +184,21 @@ func (c *NodeController) doPoolsCalculation() (map[string]string, error) {
...
-		string(rootCertBytes), path.Join(itzoDir, "ca.crt"), "0644")
-	c.CloudInitFile.AddKipFile(
-		string(certBytes), path.Join(itzoDir, "server.crt"), "0644")
-	c.CloudInitFile.AddKipFile(
-		string(keyBytes), path.Join(itzoDir, "server.key"), "0600")
-	c.CloudInitFile.AddItzoVersion(c.Config.ItzoVersion)
-	c.CloudInitFile.AddItzoURL(c.Config.ItzoURL)
-	if len(c.Config.CellConfig) > 0 {
-		c.CloudInitFile.AddCellConfig(c.Config.CellConfig)
+	if !c.Config.UseCloudParameterStore {
```
hidalgopl, line 188: minor / stylistic: I'd align default behavior of this function (which is `!c.Config.UseCloudParameterStore`) to right.
so instead having big block with default behavior in `if !var{}`, I'd change it to 
```
if c.Config.UseCloudParameterStore {
    return nil
}
... default behavior
```

pkg/server/cloud/aws/instances.go
```diff
@@ -647,12 +662,59 @@ func isUnsupportedInstanceError(err error) bool {
...
+	if n > 1 {
+		klog.Warningf("DescribeIamInstanceProfileAssociations() %s: got %d results", instanceID, n)
+	}
+
+	for _, association := range out.IamInstanceProfileAssociations {
+		if association == nil {
+			return fmt.Errorf("DescribeIamInstanceProfileAssociations() %s: nil association", instanceID)
+		}
+		klog.V(5).Infof("replacing current IAM instance profile %s on %s association ID %s state %s",
+			aws.StringValue(out.IamInstanceProfileAssociations[0].IamInstanceProfile.Arn),
```
hidalgopl, line 703: Are those log lines correct? Shouldn't we log
`aws.StringValue(association.IamInstanceProfile.Arn)` instead of always logging first element from `out.IamInstanceProfileAssociations`?

hidalgopl (approved):
good stuff :+1:

pkg/server/config.go
ldx: Good point :+1:

pkg/server/nodemanager/node_controller.go
ldx: :+1:

pkg/server/cloud/aws/instances.go
ldx: Yeah, I added a `break` at the end of the loop. The API only returns one result, but better be making sure that we're not trying to update the association more than once.

PR #195, opened 2020-12-15
Load instance data from json - closes #191

### LINKED ISSUE #191: Automate getting instance data
Currently, keeping instance data up-to-date is hard, as they are generated by bunch of [home-grown scripts](https://github.com/elotl/kip/tree/master/scripts/create_instance_data) which are run manually (and then commited to repo).

To avoid maintaining those scripts, we should use ready solution to get info and pricing about instances across various providers.
Possible solution can use [banzaicloud/cloudinfo](https://github.com/banzaicloud/cloudinfo). As a first step, we could use cloudinfo to get that data and a custom script to convert it to the format understandable for kip [example]https://github.com/elotl/kip/blob/master/pkg/util/instanceselector/aws_instance_data.go). This could be done as a part of kip's release pipeline, so we can be sure that latest kip release has fresh instance data.

cloudinfo currently supports storing data in redis or cassandra, so we probably need to scrape data, store it in redis, and then fetch it and save to json.

hidalgopl:
Added two kip flags: `override-instance-data` & `instance-data-path`. If first one is set, kip will try to load instance data from path passed to second one. If that fails it will fallback to data that is already in repo.


pkg/server/server.go
```diff
@@ -251,11 +251,17 @@ func NewInstanceProvider(configFilePath, nodeName, internalIP, clusterDNS, clust
 	)
 
 	klog.V(5).Infof("setting up instance selector")
+	if instanceDataPath != "" {
```
ldx, line 254: Do we need `override-instance-data` if setting `instance-data-path` to an empty string disables loading it anyway?

pkg/server/server.go
```diff
@@ -251,11 +251,17 @@ func NewInstanceProvider(configFilePath, nodeName, internalIP, clusterDNS, clust
...
+		_, err = os.Stat(instanceDataPath)
+		return nil, fmt.Errorf("cannot load custom instance data from path %s: %v", instanceDataPath, err)
```
ldx, line 256: Is an `if err != nil` check missing here?

pkg/util/instanceselector/instanceselector.go
```diff
@@ -137,9 +139,15 @@ func Setup(cloud, region, zone, defaultInstanceType string) error {
...
-func getSelectorData(data, regionOrZone string) ([]InstanceData, error) {
+func getSelectorData(data, regionOrZone, filepath string) ([]InstanceData, error) {
 	d := make(map[string][]InstanceData)
-	err := json.Unmarshal([]byte(data), &d)
+	fileData, err := getSelectorDataFromFile(filepath, regionOrZone)
+	if err == nil {
+		// If loading from file is successful, return data from there,
+		// if not, fallback to data from instanceselector package
+		return fileData, nil
+	}
```
ldx, line 149: We should log the error at a warning level to make it easier to debug if the user supplies an invalid file.

scripts/update_instance_data/pricing-updater/Makefile
```diff
@@ -0,0 +1,57 @@
+NAME=$(shell basename $(TOP_DIR))
+
+GIT_COMMIT=$(shell git rev-parse --short HEAD)
```
ldx, line 3: Probably does not matter too much here, but throughout kip, we use `git describe --dirty` to capture the build version.

scripts/update_instance_data/pricing-updater/Makefile
```diff
@@ -0,0 +1,57 @@
...
+
+$(NAME): $(PKG_SRC) $(CMD_SRC)
+	cd cmd/ && go build $(LDFLAGS) -o $(TOP_DIR)$(NAME)
+
+test: $(NAME)
+	go test ./...
+
+DKR=docker
+AWS_PROFILE ?= "elotl"
+AWS_REGION ?= "us-east-1"
```
ldx, line 26: These two variables are not used anymore?

pkg/util/instanceselector/instanceselector.go
hidalgopl: :+1: added

pkg/server/server.go
hidalgopl: good point, removed `override-instance-data` flag.

pkg/server/server.go
hidalgopl: nice catch, thanks ;)

ldx (approved):

PR #197, opened 2020-12-19
Remove a1 instances from instance data

ldx:
These are ARM instances, that we might want to support at some point, but right now regular x86_64 workloads won't be able to run on them.


jrroman (approved):
LGTM

hidalgopl (approved):
Ah, I missed that. Thanks for fixing :+1:

PR #199, opened 2020-12-30
Adding support for AWS Dedicated hosts

### LINKED ISSUE #198: Support for dedicated hosts
Currently Kip only supports regular on-demand and spot instances. Managing the lifecycle of dedicated hosts comes with additional steps. The goal is to support Dedicated Hosts in a transparent manner: the operator of Kip should be able to dispatch pods to dedicated hosts without manual steps (once they have an AWS quota). We might need to add either extra logic to the AWS backend that knows which instance types can only be launched as Dedicated Hosts, or require the operator to configure a node pool of Dedicated Hosts in provider.yaml.

jrroman:
Closes #198 

This will add support to allow us to use AWS Dedicated hosts


jrroman:
Rough outline of what needs to be done

https://docs.google.com/document/d/1pBtz6PRFxYye6uY5CpGT_u63YKGfM6CoiaxA80ulnGY/edit

pkg/api/types.go
```diff
@@ -131,6 +131,9 @@ type PodSpec struct {
 	// PodSpot is the policy that determines if a spot instance may be used for
 	// a Pod.
 	Spot PodSpot `json:"spot,omitempty"`
+	// PodDedicated is a policy that determines if a dedicated host instance may
+	// be used for a pod
```
ldx, line 135: Can you pls update the comment to match the name/type of the field?

pkg/api/types.go
```diff
@@ -1303,6 +1304,14 @@ type PodSpot struct {
...
 }
 
+// PodDedicated is the policy that determines if a dedicated host instance may
+// be used for a pod this is currently unimplemented but could be useful for
+// future use in dedicated host support it can more or less mimic the way that
+// PodSpot works with a custom type and "enum" like characteristics
+type PodDedicated struct {
+	// TODO
+}
+
```
ldx, line 1314: Do you think we'll use this in the near future? If not, it might be better to remove it from this change.

pkg/server/cloud/aws/instances.go
```diff
@@ -359,6 +359,130 @@ func (e *AwsEC2) StartNode(node *api.Node, image cloud.Image, metadata, iamPermi
...
+
+func (e *AwsEC2) retrieveOrAllocateHost(node *api.Node) (*string, error) {
+	describeOutput, err := e.client.DescribeHosts(&ec2.DescribeHostsInput{
+		Filter: []*ec2.Filter{
+			{
+				Name:   aws.String("state"),
+				Values: []*string{aws.String("available")},
+			}, {
+				Name:   aws.String("instance-type"),
+				Values: []*string{aws.String(node.Spec.InstanceType)},
```
ldx, line 409: There's also an `aws.StringSlice()` helper that creates a `[]*string`.

pkg/server/cloud/aws/instances.go
```diff
@@ -359,6 +359,130 @@ func (e *AwsEC2) StartNode(node *api.Node, image cloud.Image, metadata, iamPermi
...
+			},
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+	// if there is a host available to accept a new tentant return the host id
+	if len(describeOutput.Hosts) > 0 {
+		return describeOutput.Hosts[0].HostId, nil
+	}
```
ldx, line 419: When `DescribeHosts()` returns available dedicated hosts for the instance type we want to create, does it take into account if it has capacity for the new instance?

pkg/server/nodemanager/node_scaler.go
```diff
@@ -57,6 +57,15 @@ func (s *BindingNodeScaler) spotMatches(pod *api.Pod, node *api.Node) bool {
...
 }
 
+// we try test to see if Dedicated is set true for pod.Spec.Dedicated and
+// ensure that pod.Spec.Spot is not also set to true.
+func (s *BindingNodeScaler) dedicatedMatches(pod *api.Pod, node *api.Node) bool {
+	if node.Spec.Dedicated && pod.Spec.Spot.Policy == api.SpotNever {
+		return true
+	}
+	return false
+}
```
ldx, line 67: Is this function used anywhere?

ldx (commented):
Overall LGTM. I dropped a few questions, let's discuss them here or online.

pkg/server/nodemanager/node_scaler.go
hidalgopl: If it's used, it can be onliner: `return node.Spec.Dedicated && pod.Spec.Spot.Policy == api.SpotNever`

pkg/server/cloud/aws/instances.go
```diff
@@ -359,6 +359,130 @@ func (e *AwsEC2) StartNode(node *api.Node, image cloud.Image, metadata, iamPermi
...
+	for _, host := range hosts {
+		if len(host.Instances) > 0 {
+			continue
+		}
+		hostIdsForRelease = append(hostIdsForRelease, host.HostId)
+	}
+	// TODO do we need to know which hosts were successfully released vs which
+	// were unsuccessful? this runs within the gcLoop so it should clean up
+	// anything not cleaned on a next iteration
+	_, err = e.client.ReleaseHosts(&ec2.ReleaseHostsInput{
```
hidalgopl, line 380: I think we should log about unsuccessful release. Otherwise if the releasing fails and keep failing, it would be hard to debug

pkg/api/types.go
jrroman: Reasoning for this was because i was originally going to set up dedicated support like we have spot support with policies e.g `SpotAlways`, and `SpotNever`. However I dont think this is necessary and we can instead just use boolean. 

I have updated the comment.

pkg/server/cloud/aws/instances.go
jrroman: nice, thank you

pkg/server/cloud/aws/instances.go
jrroman: When you create a dedicated host you have to specify the instance type that the host is for. This will automatically provision the capacity for that instance type. So when we call the `retrieveOrAllocateHost(node *api.Node) (*string, error)` function we are asking it to only return a host of in which suits that instance type and is available to accept a new tenant.

pkg/server/nodemanager/node_scaler.go
jrroman: This function should have been placed inside of the `podMatchesNode()` function on line 67 of node_scaler.go

pkg/api/types.go
jrroman: I have removed this, we can add this at a later date if we deem it necessary.

pkg/server/nodemanager/node_scaler.go
jrroman: this has been simplified to ensure that pod.Spec.Dedicated matches node.Spec.Dedicated it has also been added to `podMatchesNode()` function where it was supposed to be called in  the first place.

pkg/server/cloud/aws/instances.go
jrroman: I have made updates to show the host id as well as the associated error if a host is unsuccessfully released. I have also added a comment stating why we should not be returning these errors

pkg/server/nodemanager/node_controller.go
```diff
@@ -269,6 +269,8 @@ func (c *NodeController) startSingleNode(node *api.Node, image cloud.Image, clou
 	)
 	if node.Spec.Spot {
 		instanceID, err = c.CloudClient.StartSpotNode(node, image, cloudInitData, c.Config.DefaultIAMPermissions)
+	} else if node.Spec.Dedicated {
```
henryprecheur, line 272: Here we could verify the `c.CloudClient` implements the `CloudClientDedicated` interface to verify this back-end supports dedicated hosts. If it doesn’t we issue a warning and fallback to `StartNode`.

I’m not sure if doing this now makes sense or not. Feel free to ignore this suggestion.

pkg/server/nodemanager/node_scaler.go
```diff
@@ -81,6 +87,10 @@ func (s *BindingNodeScaler) createNodeForPod(pod *api.Pod) *api.Node {
 	if pod.Spec.Spot.Policy == api.SpotAlways {
 		isSpotPod = true
 	}
+	isDedicatedPod := false
+	if pod.Spec.Dedicated {
+		isDedicatedPod = true
+	}
```
henryprecheur, line 93: Why not `isDedicated := pod.Spec.Dedicated`?

henryprecheur (commented):
Reviewed. Looks good.

pkg/server/cloud/mock.go
```diff
@@ -35,7 +35,9 @@ type MockCloudClient struct {
 
 	Starter             func(node *api.Node, image Image, metadata, iamPermissions string) (string, error)
 	SpotStarter         func(node *api.Node, image Image, metadata, iamPermissions string) (string, error)
+	DedicatedStarter    func(node *api.Node, image Image, metadata, iamPermissions string) (string, error)
 	Stopper             func(instanceID string) error
+	Releaser            func() error
```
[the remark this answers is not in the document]
jrroman: it is actually called on line 102:

```go
func (m *MockCloudClient) ReleaseDedicatedHosts() error {
    return m.Releaser()
} 
```

pkg/server/nodemanager/node_scaler.go
jrroman: simplified it further just setting node.Spec.Dedicated equal to pod.Spec.Dedicated

ldx (approved):

PR #200, opened 2021-01-04
Change AWS terraform configuration to work with Terraform 0.13+

henryprecheur:
Still work in progress. I did AWS, and GCE will be next.

Closes #196

Update Terraform configuration files to support 0.13+. The maintainers have deprecated a few things like referencing external variables from a destroy provisioner.

PR #201, opened 2021-01-06
Fix race condition in cleanup-vpc.sh

henryprecheur:
The issue only seems to affect Terraform with the configuration updated to support Terraform 0.13+. See PR #200 for reference.

1. The security group deletion logic in a retry loop
2. The instance deletion has been simplified a little. Instead of querying instance without the shutting-down and terminate state, it grabs the instances with the pending and running state. Then it waits for the instances to be fully deleted before retrying the loop
3. Keep the output from awscli, since there’s information that’s useful when debugging

PR #200 depends in this PR.


jrroman (approved):
lgtm 👍🏼

ldx (approved):

PR #203, opened 2021-01-08
Initial implementation of multi boot image

### LINKED ISSUE #202: Allow using AMIs built for architectures other than amd64
Right now kip can only use one AMI specified via `bootImageSpec` in provider.yaml.

However, AMIs are specific to the OS and CPU architecture. To support other architectures and OSes (ARM, mac1, etc) we need to be able to specify a boot image filter for each.

jrroman:
Closes #202 

WIP


jrroman:
very rough implementation to just see if i could get something running with multiple images specified in the `provider.yaml`. 

Example provider yaml for how I am testing:

```yaml
 16   bootImageSpec:
 17     owners: 689494258501
 18     filters: name=elotl-kip-*
 19     macfilters: name=elotl-kipmac-*
 20     armfilters: name=elotl-kiparm-*
```

pkg/server/nodemanager/node_controller.go
```diff
@@ -246,10 +256,27 @@ func (c *NodeController) startNodes(nodes []*api.Node, image cloud.Image) {
...
+		isMacImg := strings.Contains(img.Name, "elotl-kipmac-")
+		isDefaultImg := strings.Contains(img.Name, "elotl-kip-")
+		if isMacInst && isMacImg {
+			image = img
+		}
+		if !isMacInst && isDefaultImg {
+			image = img
+		}
+	}
+	return image
```
henryprecheur, line 277: We could return an “empty” image here if images aren’t named properly.

Either return an error, or a boolean indicating whether an image has been found in the list to avoid weird bugs in the future.

ldx (commented):
One thing to keep in mind is that the keys and values in the boot image spec map are cloud-specific. A more straightforward way to handle multiple images might be changing `GetImage()` to `GetImages()` that has a `BootImageSpec` parameter, and it return a `map[string]string` where the keys are architecture/OS strings e.g. `x86_64_mac`, and the values are the image IDs. Then the last piece is a map of instance family -> architecture the way `getImageForInstance()` does it.

pkg/server/nodemanager/node_controller.go
```diff
@@ -731,27 +762,33 @@ func (c *NodeController) requestNode(nodeReq NodeRequest, podNodeMapping map[str
...
+	for _, ispec := range spec.Specs() {
+		var img cloud.Image
+		obj, exists := c.ImageIdCache.Get(ispec.String())
+		if obj != nil {
+			img = obj.(cloud.Image)
+			images = append(images, img)
+		}
+		if !exists || img.ID == "" {
+			var err error
+			img, err = c.CloudClient.GetImage(ispec)
```
henryprecheur, line 778: Turn into GetImages(*specs) -> map[string]cloud.Image

henryprecheur:
Now it’s x86_64_mac_filters instead of the simpler macfilters.

We like to be specific :-)

pkg/server/cloud/cloud.go
```diff
@@ -42,6 +42,16 @@ const AWSUserIDTagKey = "AWSUserID"
...
 const InstanceParameterCertificate = "certificate"
 
+type Architecture int
+
+// Architectures aren’t represented as strings since the naming across the
+// industry isn’t uniform.
+const (
+	Arch_x86_64 Architecture = iota
+	// Macs in AWS
+	Arch_x86_64_mac
```
ldx, line 52: Can we use the regular CamelCase Go convention here, with no underscores?

pkg/server/cloud/aws/aws.go
```diff
@@ -318,3 +318,39 @@ func ec2TagsFromLabels(resource string, labels map[string]string) ([]*ec2.Tag, e
...
+		}
+		if owners_id != "" {
+			value["owners-id"] = owners_id
+		}
+		if owners_alias != "" {
+			value["owners_alias"] = owners_alias
+		}
+		specs = append(specs, value)
+	}
+	return specs
```
ldx, line 346: TAL at `bootImageSpecToDescribeImagesInput()` in `pkg/server/cloud/aws/instances.go` to see the currently supported fields for `BootImageSpec` on AWS.

pkg/server/nodemanager/node_scaler.go
```diff
@@ -84,10 +85,17 @@ func (s *BindingNodeScaler) createNodeForPod(pod *api.Pod) *api.Node {
 	if s.bootLimiter.IsUnavailableInstance(pod.Spec.InstanceType, isSpotPod) {
 		return nil
 	}
-
+	// XXX henry: we need the instance’s architecture to be able to pick the right image.
+	// This is currently handled by CloudClient, which isn’t accessible from this object.
+	// We assume x84_64 because everybody still uses it.
+	image, found := BootImages[cloud.Arch_x86_64]
```
ldx, line 91: How is this going to work for mac1 instances?

pkg/server/cloud/cloud.go
```diff
@@ -70,6 +80,14 @@ type CloudClient interface {
 	IsAvailable() (bool, error)
 	AddInstanceParameter(instanceID, key, value string, isSecret bool) error
 	DeleteInstanceParameter(instanceID, key string) error
+
+	// Because a boot image spec can match multiple architectures we turn it into
+	// a list. Refactor for later: we should probably use different types.
+	Extend(spec BootImageSpec) []BootImageSpec
```
ldx, line 86: If you look at naming in this interface, all the function names are descriptive of what they are supposed to do. `Extend` is very ambiguous, I think it'd be better to call this `SplitBootImageSpec` or something similar.

pkg/server/nodemanager/node_controller.go
```diff
@@ -246,10 +252,32 @@ func (c *NodeController) startNodes(nodes []*api.Node, image cloud.Image) {
...
+		isMacInst := strings.HasPrefix(instType, "mac")
+		if isMacInst && img.Architecture == cloud.Arch_x86_64_mac {
+			return img, true
+		}
+		if !isMacInst && img.Architecture == cloud.Arch_x86_64 {
+			return img, true
+		}
+	}
+	return cloud.Image{}, false
+}
```
ldx, line 279: I think this should go in the instanceselector package, that's the one that knows about instance families and types.

pkg/server/nodemanager/node_scaler.go
ldx: See my comment on using the instanceselector package. We can do the instance family -> architecture mapping there.

ldx (commented):
We'll also need more test cases covering the refactored `imageSpecToImage()`.

pkg/server/cloud/aws/aws.go
henryprecheur: Thanks!

pkg/api/types.go
```diff
@@ -303,6 +303,7 @@ type PodSpec struct {
 	// +optional
 	SetHostnameAsFQDN *bool `json:"setHostnameAsFQDN,omitempty"`
 
+	Architecture Architecture `json:"architecture"`
 }
```
henryprecheur, line 307: I can add methods to encode & decode the architecture type as strings if API niceties is important.

Or shall I not include it in the JSON?

pkg/util/instanceselector/instanceselector.go
```diff
@@ -450,32 +471,33 @@ func instanceTypeSpecified(instanceType string) bool {
...
 
-	instanceType, needsSustainedCPU := selector.getInstanceFromResources(ps.Resources, ps.InstanceType)
+	arch, instanceType, needsSustainedCPU := selector.getInstanceFromResources(ps.Resources, ps.InstanceType)
 	if instanceType == "" {
 		msg := "could not compute instance type from Spec.Resources. It's likely that the Pod.Spec.Resources specify an instance that doesnt exist in the cloud"
-		return "", nil, fmt.Errorf(msg)
+		return api.ArchUndefined, "", nil, fmt.Errorf(msg)
 	}
-	return instanceType, &needsSustainedCPU, nil
+	return arch, instanceType, &needsSustainedCPU, nil
```
ldx, line 500: I'm not sure if we need to handle architecture so explicitly here.

Since an instance family/type also implicitly tells you the architecture (.e.g. `mac1` is always `arch_x86_64_mac`, `c6g` is always `arm`, `m5` is always `arch_x86_64`, etc) we can simply add a function to the `instanceselector` package that gives you back the architecture of the instance family or type. I think there's only one place where we need to look it up in the node controller; this way, we would not need to sprinkle and thread through architecture as another parameter/variable in the pod registry, the instance selector, the node controller etc.

henryprecheur: This is also required in the NodeScaler, and it doesn’t have access any reference to the instance selector as far as I know. That’s why I passed it through to the PodSpec.

ldx: We can import the instanceselector package. I think it's better to keep the knowledge on architecture and how it relates to instance types confined to that package, instead of sprinkling it over a different packages.

pkg/util/instanceselector/instanceselector.go
```diff
@@ -70,10 +70,23 @@ type instanceSelector struct {
...
+	return api.ArchX8664
+}
+
+func AWSArchitectureSelector(instanceType string) api.Architecture {
+	if strings.HasPrefix(instanceType, "mac1.") {
+		return api.ArchX8664Mac
+	} else {
+		return api.ArchX8664
+	}
+}
```
henryprecheur, line 88: Replace this by querying AWS if possible.

PR #207, opened 2021-01-12
Fix my first bug: bad terminate-instances command

henryprecheur:
Because I quoted the variable instance_ids in the terminate-instance for loop, cleanup-vpc.sh doesn’t work with more than one instance. This PR fixes this by using a loop to properly expand the IDS.

I didn’t see this bug during my tests because I didn’t add any pod to the cluster.

Here’s the log of the error:

```
null_resource.gw (local-exec): An error occurred (InvalidInstanceID.Malformed) when calling the TerminateInstances operation: Invalid id: "i-026d5801edec31980
null_resource.gw (local-exec): i-0280cd2a27136b5ac"

null_resource.main (local-exec): An error occurred (InvalidInstanceID.Malformed) when calling the TerminateInstances operation: Invalid id: "i-026d5801edec31980
null_resource.main (local-exec): i-0280cd2a27136b5ac"

null_resource.gw (local-exec): Waiter InstanceTerminated failed: Invalid id: "i-026d5801edec31980
null_resource.gw (local-exec): i-0280cd2a27136b5ac"

null_resource.main (local-exec): Waiter InstanceTerminated failed: Invalid id: "i-026d5801edec31980
null_resource.main (local-exec): i-0280cd2a27136b5ac"
null_resource.main (local-exec): Terminating instances: i-026d5801edec31980
null_resource.main (local-exec): i-0280cd2a27136b5ac
null_resource.gw (local-exec): Terminating instances: i-026d5801edec31980
null_resource.gw (local-exec): i-0280cd2a27136b5ac
```

### DIFF
--- a/deploy/terraform-aws/cleanup-vpc.sh
+++ b/deploy/terraform-aws/cleanup-vpc.sh
@@ -55,9 +55,13 @@ get_pending_and_running_instance_ids_by_vpc_id() {
 instance_ids=$(get_pending_and_running_instance_ids_by_vpc_id "$VPC_ID")
 while [[ -n "$instance_ids" ]]
 do
-    echo "Terminating instances: $instance_ids"
-    aws ec2 terminate-instances --instance-ids "$instance_ids"
-    aws ec2 wait instance-terminated --instance-ids "$instance_ids"
+    for instance_id in $instance_ids
+    do
+        echo "Terminating instance: $instance_id"
+        aws ec2 terminate-instances --instance-ids "$instance_id"
+        aws ec2 wait instance-terminated --instance-ids "$instance_id"
+        echo "Instance $instance_id terminated"
+    done
 
     # Update the list to ensure nothing else came up in the VPC that would
     # block the destruction of the security group.


deploy/terraform-aws/cleanup-vpc.sh
```diff
@@ -55,9 +55,16 @@ get_pending_and_running_instance_ids_by_vpc_id() {
...
+    for instance_id in $instance_ids
+    do
+        echo "Terminating instance: $instance_id"
+        aws ec2 terminate-instances --instance-ids "$instance_id"
+    done
+    for instance_id in $instance_ids
+    do
+        aws ec2 wait instance-terminated --instance-ids "$instance_id"
+        echo "Instance $instance_id terminated"
+    done
```
henryprecheur, line 67: I made these 2 loops because AWS recommend to limit the number of IDs passed to terminate-instances.

jrroman (approved):
LGTM 👍🏼 

only had that one comment, let me know if i am missing something.

deploy/terraform-aws/cleanup-vpc.sh
```diff
@@ -55,9 +55,16 @@ get_pending_and_running_instance_ids_by_vpc_id() {
...
-    aws ec2 terminate-instances --instance-ids "$instance_ids"
-    aws ec2 wait instance-terminated --instance-ids "$instance_ids"
+    for instance_id in $instance_ids
+    do
+        echo "Terminating instance: $instance_id"
+        aws ec2 terminate-instances --instance-ids "$instance_id"
+    done
+    for instance_id in $instance_ids
+    do
+        aws ec2 wait instance-terminated --instance-ids "$instance_id"
```
jrroman, line 65: does wait halt execution until it sees that the id is in a state of terminate or does this run async and poll until the state becomes terminated?

deploy/terraform-aws/cleanup-vpc.sh
jrroman: nit: I understand what you mean however it looks like you could add `aws ec2 wait` to the first for loop. It looks like `aws ec2 wait` polls the aws describe instances api every 15 seconds until the desired state was reached. It would just mean that we wait on each iteration until state becomes terminated rather than sending the request for termination to every node then checking afterward if it completes or not.

henryprecheur: Yep we could make it a single loop. This would simplify things slightly, at the cost of more time during cluster shutdown.

I’m going to do what you suggest: given how brittle shell can be, better go for simple instead of trying to optimize for pennies of AWS credits.

PR #208, opened 2021-01-12
Add SAN to client certificate to fix Kip built with 1.15 with EC2 instances.

### LINKED ISSUE #176: Cannot ping an AWS cell using go v1.15
I built kip locally with go v1.15 and I'm mounting it as a volume, as described here: https://github.com/elotl/kip-minikube/

It starts normally, but when I tried to run nginx on virtual node, it seems that node_controller cannot (first few tries result in connection refused, but I guess it's expected as EC2 instance is booting) ping the cell:
```
W0914 14:20:52.134617       7 node_controller.go:509] Heartbeat error from node d222e754-34f0-4667-b117-2d52e1f1e2b6: Get "https://3.227.243.231:6421/rest/v1/ping": x509: certificate relies on legacy Common Name field, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0
```
It's not an issue for me right now, as I can always use `v1.13`, but I guess we need to tackle it sooner or later.

Found this discussion: https://github.com/golang/go/issues/39568

henryprecheur:
Should fix #176 
Unfortunately I couldn’t reproduce the bug with go 1.13.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+# Override the GO variable to test with a different version of Golang:
+#     $ make GO=/usr/lib/go-1.15/bin/go
+GO=go
 DKR=docker
 GIT_VERSION=$(shell git describe --dirty)
 CURRENT_TIME=$(shell date +%Y%m%d%H%M%S)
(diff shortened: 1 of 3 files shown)


henryprecheur:
I couldn’t reproduce the bug unfortunately :-( I tried with Go 1.13 and it looked like nothing changed.

I’ll need help from @hidalgopl for that.

ldx (approved):
:+1:

PR #211, opened 2021-01-15
Dedicated host support

jrroman:
These are mostly test values for debugging cloudinit issues for macos

vpc-id and subnet-id stated to pin kip to an AZ for dedicated hosts


deploy/manifests/kip/base/cloudinit.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+anka license activate -f $ANKA_LICENSE
+anka license accept-eula || true
```
hidalgopl, line 4: I believe that we need to run those anka commands with `sudo`

deploy/manifests/kip/base/cloudinit.sh
```diff
@@ -0,0 +1,9 @@
...
+anka license validate
+
+wget -O /usr/local/bin/itzo https://itzo-kip-dev-download.s3.amazonaws.com/itzo-darwin-818
```
hidalgopl, line 7: wget isn't installed, let's use curl

jrroman:
Sorry @hidalgopl this is not actually a real PR it was just reflecting the changes necessary for potentially using a secondary mac cloudinit file. More so to reflect that vilmos could see changes on my local

deploy/manifests/kip/base/cloudinit.sh
jrroman: if it were executed by mac os init it would be executed as a root user

PR #212, opened 2021-01-19
fix typo in json tag for useCloudParameterStore

hidalgopl:
just to stay in sync with docs and everything


henryprecheur (approved):

ldx (approved):
Ouch, something to keep in mind when we upgrade :+1:

PR #213, opened 2021-01-19
Dedicated Host fixes

jrroman:
**Updates:**
- Add AZ targeting support when finding dedicated hosts
- Increase boot timeout so mac instances are not prematurely purged if they take a longer period of time to boot.


ldx (approved):
:+1:

PR #217, opened 2021-01-26
check if allocated dedicated host is available (with timeout) - fixes #216

### LINKED ISSUE #216: Make sure dedicated host is "available" before launching an instance onto it

hidalgopl:
There are two issues that this should fix:
1. If there's no dedicated host available, kip tried to allocate a new one and run an instance on it without check if it's available.
2. If there was an available host but fully occupied, kip tried to run instance on it, as it was checking only `state`. Should be fixed now.

I'm still getting
```
kip-provider-0 kip I0126 14:51:01.640099       1 node_controller.go:327] Unhealthy wait for running: waiting for instance to start: ResourceNotReady: failed waiting for successful resource state, terminating node: 31ecc3c4-b4b8-47b2-9fd0-52d94514ebd8
```
from 
  `e.client.WaitUntilInstanceRunning(dii)`. I tried to increase retry timeout, but it didn't help.

### DIFF
--- a/pkg/server/cloud/aws/instances.go
+++ b/pkg/server/cloud/aws/instances.go
@@ -501,6 +501,7 @@ func (e *AwsEC2) StartDedicatedNode(node *api.Node, image cloud.Image, metadata,
 		Tags:         tags,
 	}
 	volSizeGiB := cloud.ToSaneVolumeSize(node.Spec.Resources.VolumeSize, image)
+	klog.V(2).Infof("calculated volume size for node: %v", volSizeGiB)
 	devices := e.getBlockDeviceMapping(image, volSizeGiB)
 	networkSpec := e.getInstanceNetworkSpec(node.Spec.Resources.PrivateIPOnly)
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
@@ -512,6 +513,10 @@ func (e *AwsEC2) StartDedicatedNode(node *api.Node, image cloud.Image, metadata,
 
 	klog.V(2).Infof("Starting node with security groups: %v subnet: '%s'",
 		e.bootSecurityGroupIDs, e.subnetID)
+	klog.V(2).Infof("Block devices for a node: %")
+	for _, device := range devices {
+		klog.V(2).Infof("Device: %s volume size: %d", device.DeviceName, device.Ebs.VolumeSize)
+	}
 	result, err := e.client.RunInstances(&ec2.RunInstancesInput{
 		ImageId:             aws.String(node.Spec.BootImage),
 		InstanceType:        aws.String(node.Spec.InstanceType),
--- a/pkg/server/cloud/aws/instances.go
+++ b/pkg/server/cloud/aws/instances.go
@@ -633,7 +633,9 @@ func (e *AwsEC2) WaitForRunning(node *api.Node) ([]api.NetworkAddress, error) {
 			return waitErr
 		},
 		func(err error) bool {
-			return strings.HasPrefix(err.Error(), "InvalidInstanceID.NotFound")
+			// TODO: remove this log line
+			klog.V(2).Infof("retrying err: %s", err.Error())
+			return strings.HasPrefix(err.Error(), "InvalidInstanceID.NotFound") || strings.Contains(err.Error(), "ResourceNotReady")
 		})
 	if err != nil {
 		return nil, fmt.Errorf("waiting for instance to start: %v", err)
--- a/pkg/server/cloud/aws/instances.go
+++ b/pkg/server/cloud/aws/instances.go
@@ -472,18 +472,19 @@ func (e *AwsEC2) isHostAvailable(hostId *string) bool {
 func (e *AwsEC2) waitForHostAvailable(ctx context.Context, hostId *string) bool {
 	hostAvailable := false
 	ticker := time.NewTicker(1*time.Second)
-	quit := make(chan bool)
 	for {
 		select {
 		case <-ctx.Done():
 			return hostAvailable
 		case <-ticker.C:
+			klog.V(2).Infof("checking host %s availability...", *hostId)
 			hostAvailable = e.isHostAvailable(hostId)
 			if hostAvailable {
-				quit <- true
+				return hostAvailable
 			}
-		case <-quit:
-			return hostAvailable
+			klog.V(2).Infof("host not available yet")
+		default:
+			//
 		}
 	}
 }
--- a/pkg/server/cloud/aws/instances.go
+++ b/pkg/server/cloud/aws/instances.go
@@ -86,6 +86,7 @@ func (e *AwsEC2) getNodeTags(node *api.Node) []*ec2.Tag {
 
 func (e *AwsEC2) getBlockDeviceMapping(image cloud.Image, volSizeGiB int32) []*ec2.BlockDeviceMapping {
 	awsVolSize := aws.Int64(int64(volSizeGiB))
+	klog.V(2).Infof("getting Block device mapping, got %v volSizeGiB, using %v", volSizeGiB, awsVolSize)
 	devices := []*ec2.BlockDeviceMapping{
 		&ec2.BlockDeviceMapping{
 			DeviceName: aws.String(image.RootDevice),
--- a/pkg/server/cloud/aws/instances.go
+++ b/pkg/server/cloud/aws/instances.go
@@ -463,7 +463,7 @@ func (e *AwsEC2) waitForHostAvailable(ctx context.Context, hostId *string) error
 		select {
 		case <-ctx.Done():
 			if !hostAvailable {
-				return fmt.Errorf("")
+				return fmt.Errorf("host not available")
 			} else {
 				return nil
 			}
--- a/pkg/server/cloud/aws/instances.go
+++ b/pkg/server/cloud/aws/instances.go
@@ -33,17 +33,17 @@ import (
 )
 
 const (
-	awsInstanceProduct                  = "Linux/UNIX"
-	resizeTimeout                       = 60 * time.Second
-	maxUserInstanceTags                 = 45
-	awsCreationDateFormat               = "2006-01-02T15:04:05.000Z"
-	elotlOwnerID                        = "689494258501"
-	elotlImageNameFilter                = "elotl-kip-*"
+	awsInstanceProduct    = "Linux/UNIX"
+	resizeTimeout         = 60 * time.Second
+	maxUserInstanceTags   = 45
+	awsCreationDateFormat = "2006-01-02T15:04:05.000Z"
+	elotlOwnerID          = "689494258501"
+	elotlImageNameFilter  = "elotl-kip-*"
 )
 
 var (
-	BootTimeout time.Duration = 15 * time.Minute
-	AwsInstanceAvailableState = "available"
+	BootTimeout               time.Duration = 15 * time.Minute
+	AwsInstanceAvailableState               = "available"
 )
 
 func (e *AwsEC2) StopInstance(instanceID string) error {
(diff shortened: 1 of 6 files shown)


jrroman:
> There are two issues that this should fix:
> 
>     1. If there's no dedicated host available, kip tried to allocate a new one and run an instance on it without check if it's available.
> 
>     2. If there was an available host but fully occupied, kip tried to run instance on it, as it was checking only `state`. Should be fixed now.

I don't know that 2 is actually an issue within the implementation of requesting and allocating hosts we are only returning hosts that are not currently occupied. Im not sure as to where you are seeing that we are only checking for 'state'?

pkg/server/cloud/aws/instances.go
```diff
@@ -17,7 +17,9 @@ limitations under the License.
 package aws
 
 import (
+	"context"
 	"fmt"
+	"github.com/elotl/kip/pkg/server/nodemanager"
```
ldx, line 22: This is a no-no: even though it's not an explicit circular dependency, the node controller uses the cloud client which is implemented by this package for AWS, so it's pretty close. It's 100x better to hardcode a timeout value than to introduce a circular dependency.

pkg/server/cloud/aws/instances.go
```diff
@@ -455,6 +496,12 @@ func (e *AwsEC2) StartDedicatedNode(node *api.Node, image cloud.Image, metadata,
 	volSizeGiB := cloud.ToSaneVolumeSize(node.Spec.Resources.VolumeSize, image)
 	devices := e.getBlockDeviceMapping(image, volSizeGiB)
 	networkSpec := e.getInstanceNetworkSpec(node.Spec.Resources.PrivateIPOnly)
+	ctx, _ := context.WithTimeout(context.Background(), 30 * time.Second)
```
ldx, line 499: The second return value should be called before returning this function to prevent a resource leak.

pkg/server/cloud/aws/instances.go
```diff
@@ -441,6 +450,38 @@ func (e *AwsEC2) retrieveOrAllocateHost(node *api.Node) (*string, error) {
...
+	for {
+		select {
+		case <-ctx.Done():
+			if !hostAvailable {
+				return fmt.Errorf("host not available")
+			} else {
+				return nil
+			}
+		default:
+			hostAvailable = e.isHostAvailable(hostId)
```
ldx, line 479: Should not this return once the host is available, to prevent unnecessary waiting?

pkg/server/cloud/aws/instances.go
```diff
@@ -441,6 +450,38 @@ func (e *AwsEC2) retrieveOrAllocateHost(node *api.Node) (*string, error) {
...
+		}
+		time.Sleep(500 * time.Millisecond)
```
ldx, line 481: Instead of a direct sleep here, you could use a Ticker in the `select` statement above instead of `default`.

pkg/server/cloud/aws/instances.go
```diff
@@ -441,6 +450,38 @@ func (e *AwsEC2) retrieveOrAllocateHost(node *api.Node) (*string, error) {
...
+	})
+	if err != nil {
+		klog.Errorf("cannot describe host %s: %v", *hostId, err)
+		return false
+	}
+	if len(describeOutput.Hosts) != 1 {
+		return false
+	}
+	host := describeOutput.Hosts[0]
+	return *host.State == "available"
```
ldx, line 465: Please use `aws.StringValue()` to check the value safely.

pkg/server/cloud/aws/instances.go
```diff
@@ -441,6 +450,38 @@ func (e *AwsEC2) retrieveOrAllocateHost(node *api.Node) (*string, error) {
 	return allocateOutput.HostIds[0], nil
 }
 
+func (e *AwsEC2) isHostAvailable(hostId *string) bool {
+	describeOutput, err := e.client.DescribeHosts(&ec2.DescribeHostsInput{
+		HostIds:    []*string{hostId},
```
ldx, line 455: The `aws` package has a function for create a slice of string pointers, IIRC `aws.StringSlice()`.

ldx (commented):
Thanks for fixing this! I've dropped a few comments with suggestions.

pkg/server/cloud/aws/instances.go
hidalgopl: good point :+1: I forgot to copy this var to the aws package.

hidalgopl:
FYI
I think that the issue we're hitting is related with volume size of the AMI. We started seeing it after we resized attached volume to 500Gb. After my changes from this PR, kip gets to the point when it sends `RunInstances` request, but instance creation fails and dedicated host almost immediately turns its state to Pending.
Screenshot from 2021-01-27 19-13-32
As you see, there's `Client.InvalidParameterCombination: Could not create volume with size 500GiB from snapshot 'snap-064acde5cd376cb6b'`. I've added logging block devices volume size to debug if correct numbers are passed to `RunInstances` function call.
```
kip-provider-0 kip I0127 17:45:33.909398       1 instances.go:493] Starting instance for node: &{{Node v1} {26b5749b-a378-4a77-9ece-05314c889606 map[] 2021-01-27 17:45:33.90770785 +0000 UTC <nil> map[] 08358c1e-5551-4f58-be7a-c4119665a587 default} {mac1.metal ami-0929815870cdeaa46 false false true {   20G false <nil> false <nil>}} {Creating  [] default_buildkite-agent-mac1-metal}}
kip-provider-0 kip I0127 17:45:34.209321       1 instances.go:504] calculated volume size for node: 500
kip-provider-0 kip I0127 17:45:35.209502       1 instances.go:480] checking host h-020fba8d6384e6cc9 availability...
kip-provider-0 kip I0127 17:45:35.281221       1 instances.go:514] Starting node with security groups: [sg-0ee086454488a7451] subnet: 'subnet-2769f140'
kip-provider-0 kip I0127 17:45:35.281250       1 instances.go:516] Block devices for a node
kip-provider-0 kip I0127 17:45:35.281257       1 instances.go:518] Device: /dev/sda1 volume size: 824658735208
kip-provider-0 kip I0127 17:45:36.707965       1 instances.go:552] Started instance: i-01b2507ae043e4275
kip-provider-0 kip I0127 17:45:51.843890       1 instances.go:642] retrying err: ResourceNotReady: failed waiting for successful resource state
``` 
there's probably an issue with the volume size, as you may see from logs

pkg/server/cloud/aws/instances.go
```diff
@@ -462,23 +466,25 @@ func (e *AwsEC2) isHostAvailable(hostId *string) bool {
...
-			}
-		default:
+			return hostAvailable
+		case <-ticker.C:
 			hostAvailable = e.isHostAvailable(hostId)
+			if hostAvailable {
+				quit <- true
+			}
+		case <-quit:
+			return hostAvailable
```
ldx: This could be simplified further if you removed `quit`, and return `hostAvailable` directly when it's true after the `isHostAvailable()` check.

ldx (approved):

pkg/server/nodemanager/node_scaler.go
```diff
@@ -102,8 +102,9 @@ func (s *BindingNodeScaler) createNodeForPod(pod *api.Pod) *api.Node {
 	// until dispatch (just launch with the default size), otherwise,
 	// if we have a fixed size disk of non-default size, size the disk
 	// here and also take care of matching disk sizes
+
 	node.Spec.Resources.VolumeSize = s.defaultVolumeSize
-	if s.fixedSizeVolume && pod.Spec.Resources.VolumeSize != "" {
+	if pod.Spec.Resources.VolumeSize != "" { // TODO: consider if && s.fixedSizeVolume is needed here
```
hidalgopl, line 107: @ldx @henryprecheur for AWS `s.fixedSizeVolume` is always set to false. Which basically means that pod annotation `pod.elotl.co/volume-size: "X Gi"` has no effect, which may be confusing for user (it isn't also specified nowhere in docs for which cloud providers this annotation works). I needed to manually set volume size to test if the size is an issue and noticed that. Should we address it in a separate issue or is this desired behavior?

hidalgopl:
Ok, so there were few issues here:
1. Retrying timeout for checking if instance is available was too small - I increased it 
2. We had hardcoded volume type for all EBSes - `gp2`. I've added getting volume type from AMI's volumes and using it.
3. We were checking only for host state (if it's available) without checking if it has free compute (which result in attempts of trying to run an instance on the occupied host) - I've added additional check if it has available vCPUs.
4. We haven't been checking newly allocated host state - added additional check.

pkg/server/nodemanager/node_scaler.go
ldx: This annotation does work on AWS. However, resizing happens in the pod controller, not in the node controller when `fixedSizeVolume` is false (i.e. on AWS).

pkg/server/cloud/aws/instances.go
```diff
@@ -441,6 +466,41 @@ func (e *AwsEC2) retrieveOrAllocateHost(node *api.Node) (*string, error) {
...
+		select {
+		case <-ctx.Done():
+			return hostAvailable
+		case <-ticker.C:
+			hostAvailable = e.isHostAvailable(hostId)
+			if hostAvailable {
+				return hostAvailable
+			}
+			klog.V(2).Infof("host %s not available yet", *hostId)
+		default:
```
ldx, line 498: Don't leave in the `default` case, otherwise it will always be taken while the other two cases are not true, effectively busy looping here.

pkg/server/nodemanager/node_controller.go
```diff
@@ -324,7 +324,6 @@ func (c *NodeController) finishNodeStart(node *api.Node) error {
 	// a describe instance here...
 	addresses, err := c.CloudClient.WaitForRunning(node)
 	if err != nil {
-		klog.V(2).Infof("Unhealthy wait for running, terminating node: %s", node.Name)
```
ldx, line 327: Just double-checking that you wanted to remove this log line and we don't need it for diagnosing node start problems?

ldx (approved):
:+1:

pkg/server/cloud/aws/instances.go
```diff
@@ -441,6 +466,41 @@ func (e *AwsEC2) retrieveOrAllocateHost(node *api.Node) (*string, error) {
...
+	}
+	if len(describeOutput.Hosts) != 1 {
+		klog.Errorf("no such host: %s", aws.StringValue(hostId))
+		return false
+	}
+	host := describeOutput.Hosts[0]
+	return aws.StringValue(host.State) == AwsInstanceAvailableState
+}
+
+func (e *AwsEC2) waitForHostAvailable(ctx context.Context, hostId *string) bool {
```
henryprecheur, line 485: This function needs a timeout. Otherwise it may wait for the host to come up forever.

pkg/server/cloud/aws/instances.go
hidalgopl: :+1: added explicit `cancel()` before sending `RunInstance` request

pkg/server/cloud/aws/instances.go
hidalgopl: :+1: removed

pkg/server/nodemanager/node_controller.go
hidalgopl: Good catch! I went too far in removing debug lines added by me :+1: reverted

pkg/server/cloud/aws/instances.go
```diff
@@ -40,6 +41,11 @@ const (
 	elotlImageNameFilter  = "elotl-kip-*"
 )
 
+var (
```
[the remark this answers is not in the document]
hidalgopl: moved to `consts` block :+1:

pkg/server/cloud/aws/instances.go
```diff
@@ -274,6 +280,16 @@ func getRootDeviceVolumeSize(blockDevices []*ec2.BlockDeviceMapping, rootDeviceN
...
 
+func getDeviceVolumeType(blockDevices []*ec2.BlockDeviceMapping, rootDeviceName string) string {
+	volumeType := "gp2"
+	for _, blockDevice := range blockDevices {
+		if aws.StringValue(blockDevice.DeviceName) == rootDeviceName && blockDevice.Ebs != nil {
+			volumeType = aws.StringValue(blockDevice.Ebs.VolumeType)
+		}
+	}
+	return volumeType
+}
```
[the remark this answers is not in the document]
hidalgopl: I like the idea, merged those two :+1:

pkg/server/cloud/aws/instances.go
ldx: > This function needs a timeout. Otherwise it may wait for the host to come up forever.

The caller specifies a timeout via `ctx`.

PR #221, opened 2021-02-02
Instance selector back-pressure - fixes #214

### LINKED ISSUE #214: Implement instance selector back-pressure
Occasionally Availability Zones run out of instance types:

```
E0115 23:01:14.732148       1 node_controller.go:277] Error in node start: Region has no capacity: InsufficientInstanceCapacity : We currently do not have sufficient x1e.2xlarge capacity in the Availability Zone you requested (us-west-2b). Our system will be working on provisioning additional capacity. You can currently get x1e.2xlarge capacity by not specifying an Availability Z one in your request or choosing us-west-2a, us-west-2c.
```

The instance selector is too rigid, it picks the instance type based solely on cpu+memory needs and instance prices, it'll keep trying to start the same one even when it is out of capacity ATM in the AZ.

The cluster autoscaler does not handle this situation gracefully, you are expected to max instance types/families in the same AZ to prevent this from happening.

Kip can either:

1. Apply back-pressure by channeling back the request to the instance selector
2. Retry with a different instance type

Both of these things should happen automatically with as little manual intervention from the user/customer.

Original thread:
https://elotl.slack.com/archives/C6YDX3G1Y/p1610752350059800

hidalgopl:


pkg/util/instanceselector/instanceselector.go
```diff
@@ -344,7 +344,7 @@ func toInstanceData(data []CustomInstanceData, memoryRequirement, cpuRequirement
 // the t2.Unlimited option from AWS. For T2 instances, we try to
 // figure out what percentage of a CPU a user will likely use and
 // use that to compute t2.Unlimited cost.
-func (instSel *instanceSelector) getInstanceFromResources(rs api.ResourceSpec, instanceTypeGlob string) (string, bool) {
+func (instSel *instanceSelector) getInstanceFromResources(rs api.ResourceSpec, additionalFilter func(inst InstanceData) bool) (string, bool) {
```
henryprecheur, line 347: I like that change, but it caused this snippet of code to be duplicated:

```golang
func(inst InstanceData) bool {
	// do any additional filtering that is needed
		if instanceTypeGlob == "" {
	matches = filterInstanceData(matches, additionalFilter)
			return true
		}
		return glob.Glob(instanceTypeGlob, inst.InstanceType)
	}
```

I suggest you add and use a helper function like this to remove the duplication:

```golang
func GlobInstanceType(string) func(InstanceData) bool
```

pkg/server/nodemanager/node_controller.go
```diff
@@ -263,7 +266,7 @@ func (c *NodeController) handleStartNodeError(node *api.Node, err error, isSpot
 	}
 }
 
-func (c *NodeController) startSingleNode(node *api.Node, image cloud.Image, cloudInitData string) error {
+func (c *NodeController) startSingleNode(node *api.Node, image cloud.Image, cloudInitData string, excludeInstances []string) error {
```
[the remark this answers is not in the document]
hidalgopl: I like the idea of having a for loop instead of recursion :+1:

pkg/server/nodemanager/node_controller.go
```diff
@@ -267,13 +270,45 @@ func (c *NodeController) startSingleNode(node *api.Node, image cloud.Image, clou
...
 		err        error
+		excludeInstances []string
 	)
-	if node.Spec.Spot {
-		instanceID, err = c.CloudClient.StartSpotNode(node, image, cloudInitData, c.Config.DefaultIAMPermissions)
-	} else if node.Spec.Dedicated {
-		instanceID, err = c.CloudClient.StartDedicatedNode(node, image, cloudInitData, c.Config.DefaultIAMPermissions)
-	} else {
-		instanceID, err = c.CloudClient.StartNode(node, image, cloudInitData, c.Config.DefaultIAMPermissions)
+	for retries := 0; retries < NoCapacityRetryLimit; retries++ {
```
ldx, line 275: What's the purpose of retrying here? That will be taken care of by the control loop.

pkg/server/nodemanager/node_controller.go
```diff
@@ -267,13 +270,45 @@ func (c *NodeController) startSingleNode(node *api.Node, image cloud.Image, clou
...
+						return false
+					}
+				}
+				return true
+			})
+			klog.V(2).Infof("No sufficient capacity in region to boot %s , trying to boot %s", node.Spec.InstanceType, instanceType)
+			node.Spec.InstanceType = instanceType
+		} else {
+			break
+		}
```
ldx, line 311: We already have `pkg/server/nodemanager/boot_limiter.go` for keeping track of unavailable instances. See https://github.com/elotl/kip/blob/master/pkg/server/nodemanager/node_scaler.go#L85-L111 for a check where we're already using it to limit boot requests for unavailable instance types.

ldx (changes requested):
This would be a lot more elegant if we used `BootLimiter`, and moved the fallback logic there.

pkg/util/instanceselector/instanceselector.go
```diff
@@ -470,14 +466,27 @@ func ResourcesToInstanceType(ps *api.PodSpec) (string, *bool, error) {
...
-	instanceType, needsSustainedCPU := selector.getInstanceFromResources(ps.Resources, ps.InstanceType)
+	instanceType, needsSustainedCPU := selector.getInstanceFromResources(ps.Resources, globInstanceType(ps.InstanceType))
 	if instanceType == "" {
 		msg := "could not compute instance type from Spec.Resources. It's likely that the Pod.Spec.Resources specify an instance that doesnt exist in the cloud"
 		return "", nil, fmt.Errorf(msg)
 	}
 	return instanceType, &needsSustainedCPU, nil
 }
 
+func globInstanceType(instanceType string) func(inst InstanceData) bool {
```
ldx, line 477: I had to stare at this for a while to figure out what's going on. How about calling it `makeInstanceTypeGlobberFunc` or something similar so the reader has an easier time figuring out the logic here?

pkg/server/nodemanager/node_controller.go
```diff
@@ -275,6 +280,23 @@ func (c *NodeController) startSingleNode(node *api.Node, image cloud.Image, clou
...
+			// so let's make sure here that:
+			// 1. Selected instance type isn't the one that just failed
+			// 2. Selected instance type isn't already marked as Unavailable
+			return !c.BootLimiter.IsUnavailableInstance(inst.InstanceType, node.Spec.Spot) && !(inst.InstanceType == node.Spec.InstanceType)
+		})
+		err = &cloud.InsufficientCapacityError{
+			InstanceType:  node.Spec.InstanceType,
+			OriginalError: err.Error(),
+		}
+		node.Spec.InstanceType = instanceType
```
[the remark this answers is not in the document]
hidalgopl: added saving to handleNodeStartError :+1:

pkg/server/cloud/errors.go
```diff
@@ -24,6 +24,7 @@ type NoCapacityError struct {
 	OriginalError string
 	AZ            string
 	SubnetID      string
+	InstanceType  string
```
ldx, line 27: Is this left over? This field does not seem to be referenced anywhere in this PR.

pkg/server/nodemanager/node_controller.go
```diff
@@ -260,6 +263,11 @@ func (c *NodeController) handleStartNodeError(node *api.Node, err error, isSpot
 		// with this but I hesitate to do that, instead lets push that
 		// off to the operator for now.
 		c.BootLimiter.AddUnavailableInstance(node.Spec.InstanceType, isSpot)
+	case *cloud.InsufficientCapacityError:
+		e := err.(*cloud.InsufficientCapacityError)
+		c.BootLimiter.AddUnavailableInstance(e.InstanceType, isSpot)
+		// node has already new instance type chosen, persist it in registry
+		c.NodeRegistry.UpdateNode(node)
```
ldx, line 270: Let's check if `UpdateNode()` returned an error here, and log it.

ldx (approved):
Looks good :+1: I dropped two suggestions with some minor changes.

pkg/server/cloud/errors.go
hidalgopl: nice catch :+1:

PR #223, opened 2021-02-05
get volume throughput and iops from ami

hidalgopl:
add it to the requested block device mapping

### DIFF
--- a/pkg/server/cloud/aws/instances.go
+++ b/pkg/server/cloud/aws/instances.go
@@ -44,6 +44,13 @@ const (
 	AvailableWaitTimeout      = 30 * time.Second
 )
 
+type EbsSpecs struct {
+	Iops       int64
+	Throughput int64
+	VolumeType string
+	VolumeSize int32
+}
+
 func (e *AwsEC2) StopInstance(instanceID string) error {
 	awsInstanceIDs := []*string{aws.String(instanceID)}
 	_, err := e.client.TerminateInstances(&ec2.TerminateInstancesInput{
@@ -85,9 +92,11 @@ func (e *AwsEC2) getNodeTags(node *api.Node) []*ec2.Tag {
 func (e *AwsEC2) getBlockDeviceMapping(image cloud.Image, volSizeGiB int32) []*ec2.BlockDeviceMapping {
 	awsVolSize := aws.Int64(int64(volSizeGiB))
 	devices := []*ec2.BlockDeviceMapping{
-		&ec2.BlockDeviceMapping{
+		{
 			DeviceName: aws.String(image.RootDevice),
 			Ebs: &ec2.EbsBlockDevice{
+				Iops:                aws.Int64(image.VolumeIops),
+				Throughput:          aws.Int64(image.VolumeThroughput),
 				VolumeType:          aws.String(image.VolumeType),
 				DeleteOnTermination: aws.Bool(true),
 				VolumeSize:          awsVolSize,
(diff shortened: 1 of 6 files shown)


pkg/server/nodemanager/node_controller.go
```diff
@@ -43,7 +43,7 @@ import (
 // non-const timeouts were endorsed by Mitchell Hashimoto
 var (
 	//	BootTimeout         time.Duration = 300 * time.Second
-	BootTimeout         time.Duration = 15 * time.Minute
+	BootTimeout         time.Duration = 20 * time.Minute
```
henryprecheur, line 46: We probably should have different timeouts per instance types.

This is fine for now.

henryprecheur (commented):
Looks good to me.

I’ll let @ldx approve that.

pkg/server/cloud/aws/instances.go
```diff
@@ -44,6 +44,13 @@ const (
 	AvailableWaitTimeout      = 30 * time.Second
 )
 
+type EbsSpecs struct {
+	Iops       int64
+	Throughput int64
```
ldx, line 49: We should change these two to be pointers, since in some cases, they are optional (see below).

pkg/server/cloud/aws/instances.go
```diff
@@ -267,17 +283,20 @@ func bootImageSpecToDescribeImagesInput(spec cloud.BootImageSpec) *ec2.DescribeI
...
-	volumeType := "gp2"
+func getRootDeviceVolumeSpecs(blockDevices []*ec2.BlockDeviceMapping, rootDeviceName string) EbsSpecs {
+	specs := EbsSpecs{VolumeType: "gp2"}
 	for _, blockDevice := range blockDevices {
 		if aws.StringValue(blockDevice.DeviceName) == rootDeviceName && blockDevice.Ebs != nil {
-			rootDiskSize = int32(aws.Int64Value(blockDevice.Ebs.VolumeSize))
-			volumeType = aws.StringValue(blockDevice.Ebs.VolumeType)
+			specs.VolumeSize = int32(aws.Int64Value(blockDevice.Ebs.VolumeSize))
+			specs.VolumeType = aws.StringValue(blockDevice.Ebs.VolumeType)
+			specs.Iops = aws.Int64Value(blockDevice.Ebs.Iops)
```
ldx, line 292: `Iops` is optional  for gp3:

    	// This parameter is required for io1 and io2 volumes. The default for gp3 volumes
	// is 3,000 IOPS. This parameter is not supported for gp2, st1, sc1, or standard

So it should be a pointer in `Image`. Otherwise we'd set it to `0` in `getBlockDeviceMapping()` if it's not specified here.

pkg/server/cloud/aws/instances.go
```diff
@@ -267,17 +283,20 @@ func bootImageSpecToDescribeImagesInput(spec cloud.BootImageSpec) *ec2.DescribeI
...
+			if specs.VolumeType == "gp3" {
+				specs.Throughput = aws.Int64Value(blockDevice.Ebs.Throughput)
```
ldx, line 294: The AWS docs only say that this is only a valid parameter for gp3, but they don't mention if it is optional. I'd also make this one a pointer, and only set it in `getBlockDeviceMapping()` if it's non-nil here.

pkg/server/cloud/cloud.go
```diff
@@ -112,12 +112,14 @@ type SubnetAttributes struct {
...
-	VolumeDiskSize int32 // in GiB
-	VolumeType 	   string
+	ID               string
+	Name             string
+	RootDevice       string
+	CreationTime     *time.Time
+	VolumeDiskSize   int32 // in GiB
+	VolumeType       string
+	VolumeIops       int64
+	VolumeThroughput int64
```
ldx, line 122: Same here for `VolumeIops` and `VolumeThroughput`, it's better to make them pointers.

pkg/server/cloud/aws/instances.go
```diff
@@ -44,6 +44,13 @@ const (
 	AvailableWaitTimeout      = 30 * time.Second
 )
 
+type EbsSpecs struct {
+	Iops       int64
+	Throughput int64
+	VolumeType string
+	VolumeSize int32
```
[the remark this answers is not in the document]
hidalgopl: good point :+1: added

pkg/server/cloud/aws/instances_test.go
```diff
@@ -78,47 +78,60 @@ func TestGetRootDeviceVolumeSize(t *testing.T) {
...
 					},
 				},
 			},
-			rootDeviceName,
-			expectedVolumeSize,
+			rootDeviceName:       rootDeviceName,
+			expectedRootDiskSize: expectedVolumeSize,
+			expectedVolumeType:   "gp3",
+			expectedIops:         100,
+			expectedThroughput:   100,
```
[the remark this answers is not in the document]
hidalgopl: :+1: added

pkg/server/cloud/cloud.go
hidalgopl: changed `EbsSpec` & `cloud.Image` Throughput & Iops for pointers :+1:

ldx (approved):

PR #230, opened 2021-03-12
Require setting resourceGroup in provider.yaml for Azure fixes #228

### LINKED ISSUE #228: Fix parsing virtualNetwork name for azure
Currently, we don't return an error if `cloud.azure.virtualNetworkName` is not in required format `resource-group-name/vnet-name`.
It fails later with 404 when it tries to fetch virtual network details. We should either clarify required format in `cloud.azure.virtualNetworkName` or have another required field in config `cloud.azure.resourceGroup`. Anyway, we should return verbose error if any of these is missing.

hidalgopl:


ldx (commented):
What's the problem this change is solving?

hidalgopl:
> What's the problem this change is solving?

Before we required virtualNetwork name in format resourceGroup/vNetworkName, but it wasn't documented anywhere. I have gut feeling that we'll need resourceGroup anyway for other azure api requests, so I think it's better to require setting it separately by end user.

ldx (approved):

PR #236, opened 2021-04-21
Fix kip smoke test: use docker-ce instead of docker.io in node.sh

hidalgopl:
Our `/terraform-aws` setup start failing due to the [docker bug](https://github.com/moby/moby/issues/41890). To fix it, I switched `docker.io` (which is no longer recommened package AFAIC) to `docker-ce` based on install instruction from [official docker docs](https://docs.docker.com/engine/install/ubuntu/#install-using-the-repository). Also fixed deprecation warning about api version of`InitConfiguration` in `kubeadm init`.

### DIFF
--- a/deploy/terraform-aws/node.sh
+++ b/deploy/terraform-aws/node.sh
@@ -3,7 +3,7 @@
 set -e
 
 apt-get update
-apt-get install apt-transport-https ca-certificates curl gnupg lsb-release
+apt-get install -y apt-transport-https ca-certificates curl gnupg lsb-release
 curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
 cat <<EOF > /etc/apt/sources.list.d/kubernetes.list
 deb http://apt.kubernetes.io/ kubernetes-xenial main
--- a/deploy/terraform-aws/node.sh
+++ b/deploy/terraform-aws/node.sh
@@ -1,11 +1,16 @@
 #!/bin/bash
 
+apt-get update
+apt-get install apt-transport-https ca-certificates curl gnupg lsb-release
 curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
 cat <<EOF > /etc/apt/sources.list.d/kubernetes.list
 deb http://apt.kubernetes.io/ kubernetes-xenial main
 EOF
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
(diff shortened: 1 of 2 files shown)


deploy/terraform-aws/node.sh
```diff
@@ -1,11 +1,16 @@
 #!/bin/bash
```
henryprecheur, line 2: Use `set -e` here to exit the script when there's an error. This makes debugging the script much easier.

deploy/terraform-aws/node.sh
```diff
@@ -68,7 +73,9 @@ controllerManager:
     address: 0.0.0.0
 kubernetesVersion: "$k8s_version"
 EOF
+kubeadm config migrate --old-config /tmp/old-kubeadm-config.yaml --new-config /tmp/kubeadm-config.yaml
```
henryprecheur, line 76: Unless there's a reason to keep the old version of the file around I would just update /tmp/kubeadm-config.yaml to have the updated yaml. I don't think we should keep the old configuration around.

deploy/terraform-aws/node.sh
```diff
@@ -68,7 +73,9 @@ controllerManager:
...
 kubeadm init --config=/tmp/kubeadm-config.yaml
+stat /etc/kubernetes/admin.conf
```
henryprecheur, line 78: I think you can remove this when `set -e` is set at the top of the script. It kubeadm can't create the configuration it will certainly exit with a non-zero error code.

scripts/run_smoke_test.sh
```diff
@@ -99,7 +99,7 @@ run_smoke_test() {
     kubectl run nginx --image=nginx --port=80
     kubectl expose pod nginx
     kubectl run test --restart=Never --image=elotl/debug --command -- /bin/sh -c "$curlcmd"
-    timeout 300s bash -c "$waitcmd"
+    timeout 360s bash -c "$waitcmd"
```
henryprecheur, line 102: If 300 seconds wasn't enough and made the test flaky, I would double it to 600 seconds to ensure the flakes won't come back to haunt us.

henryprecheur (approved):

deploy/terraform-aws/node.sh
hidalgopl: Hmm, I would prefer to leave it as it is, given the fact we want to support multiple k8s versions (and k8s version could be set in terraform), so I'd prefer to let kubeadm make a decision whether config should be migrated or not

scripts/run_smoke_test.sh
hidalgopl: I set it to 420s. It should be more than sufficient; if it is exceeded there's clearly something wrong. I don't want to bump it to 10 mins as those builds are already ~20-25 mins long

PR #239, opened 2021-06-15
add signerName to CSR - closes #235

### LINKED ISSUE #235: KIP init-cert does not work on EKS 1.19
Kip uses [init-cert image](https://github.com/elotl/kip/tree/master/scripts/init-cert) as initContainer to get a certificate that allows communication with API server.

init-cert image creates certificates.k8s.io CertificateSigningRequest object and approves it using kubectl; control plane should issue a certificate and update this object status.certificate. However, it seems to be broken in EKS 1.19: after CSR approval, `status.certificate` is still nil. [There have been some changes](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html) in CertificateSigningRequest API in 1.19; we should support them while also maintaining backward compatibility.

hidalgopl:
From EKS docs:

> The CertificateSigningRequest API has been promoted to stable certificates.k8s.io/v1 with the following changes:
spec.signerName is now required. You can't create requests for kubernetes.io/legacy-unknown with the certificates.k8s.io/v1 API.
You can continue to create CSRs with the kubernetes.io/legacy-unknown signer name with the certificates.k8s.io/v1beta1 API.
You can continue to request that a CSR to is signed for a non-node server cert, webhooks, for example, with the certificates.k8s.io/v1beta1 API. These CSRs aren't auto-approved.

I added `signerName: kubernetes.io/legacy-unknown`. Once kubernetes releases v1.22.0 we should consider switching to certificates.k8s.io/v1 API. 
Bumped Kubernetes version in the CI to v1.19.0 as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
   - AWS_REGION=us-east-1
   - AWS_DEFAULT_REGION=us-east-1
   - TERRAFORM_VERSION=0.12.21
-  - KUBECTL_VERSION=1.18.0
+  - KUBECTL_VERSION=1.19.0
   - KUSTOMIZE_VERSION=3.6.1
   - GOOGLE_APPLICATION_CREDENTIALS=.account.json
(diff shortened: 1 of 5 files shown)


sunstick (approved):

henryprecheur (approved):
Looks good

PR #242, opened 2021-07-12
Add EKS kustomize overlay and EKS smoke testing

hidalgopl:
EKS smoke tests scenario:
1. They run in our EKS elotl-ci-cd cluster.
2. We deploy newly built kip version there to `kip-smoke-tests` namespace and we taint a node, so only build jobs can run on it.
3. We deploy two nginx pods: one running on KIP, one on regular node.
4. We deploy another pod which curls those two.
5. If we get expected response, tests are complete.
Those test are meant to check if:
1. Kip starts in EKS cluster
2. KIP pods start successfully.
3. Connectivity between regular <-> kip and kip <-> kip pods

Added also a README with instructions for configuring EKS cluster for KIP.

If we keep `elotl-ci-cd` on Kubernetes version which is currently default for EKS, we can be sure that we support at least current default EKS version.

### DIFF
--- a/scripts/run_eks_smoke_test.sh
+++ b/scripts/run_eks_smoke_test.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+BUILD=${TRAVIS_BUILD_NUMBER:-local}
+USE_REGION=${USE_REGION:-us-east-1}
+
+handle_error() {
+    trap - EXIT
+    show_kube_info || true
+    cleanup || true
+}
+
+cleanup() {
+    delete_kube_resources
+}
+
+delete_kube_resources() {
+    kubectl delete -f $SCRIPT_DIR/eks-smoke-test/kip-nginx.yaml
+    kubectl delete -f $SCRIPT_DIR/eks-smoke-test/regular-node-nginx.yaml
+    kubectl delete pod test
+}
+
+show_kube_info() {
+    kubectl get pods -A
+    kubectl get nodes
+    kubectl -n kip-smoke-tests describe pod -l statefulset.kubernetes.io/pod-name=kip-build-kip-provider-0
+    kubectl logs -n kip-smoke-tests -l statefulset.kubernetes.io/pod-name=kip-build-kip-provider-0r -c init-cert --tail=-1
+    kubectl logs -n kip-smoke-tests -l statefulset.kubernetes.io/pod-name=kip-build-kip-provider-0 -c kip --tail=-1
+}
+
+update_vk() {
+    local version="$(git describe)"
+    local patch_init="{\"spec\":{\"template\":{\"spec\":{\"initContainers\":[{\"image\":\"elotl/init-cert:$version\",\"name\":\"init-cert\"}]}}}}"
+    local patch_kip="{\"spec\":{\"template\":{\"spec\":{\"containers\":[{\"image\":\"elotl/kip:$version\",\"name\":\"kip\"}]}}}}"
+    kubectl patch -n kip-smoke-tests statefulset kip-build-kip-provider -p "$patch_init"
+    kubectl patch -n kip-smoke-tests statefulset kip-build-kip-provider -p "$patch_kip"
+}
+
+run_smoke_test_1() {
+     kubectl apply -f $SCRIPT_DIR/eks-smoke-test/kip-nginx.yaml
+     kubectl apply -f $SCRIPT_DIR/eks-smoke-test/regular-node-nginx.yaml
+     kubectl get nodes -o wide
+     kubectl get pods -n kip-smoke-tests
+#    local curlcmd="i=0; while [ \$i -lt 300 ]; do i=\$((i+1)); curl nginx | grep 'Welcome to nginx' && exit 0; sleep 1; done; exit 1"
+#    local waitcmd="phase=\"\"; echo \"Waiting for test results from pod\"; until [[ \$phase = Succeeded ]]; do sleep 1; phase=\$(kubectl get pod test -ojsonpath=\"{.status.phase}\"); if [[ \$phase = Failed ]]; then echo \$phase; kubectl get pods -A; exit 1; fi; echo \$phase; done"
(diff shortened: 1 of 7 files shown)


scripts/run_eks_smoke_test.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
henryprecheur, line 3: I would make this an explicit parameter. Changing the behavior of the script depending of where it's invoked from is kind of a gotcha.

Invoke it like this:
```lang=shell
$ ./scripts/run_eks_smoke_test.sh ./scripts/eks-smoke/test
```
And in this script you'd have:
```lang=shell
if [ $# -neq 1 ]
then
    echo "Usage: $0 path/to/yaml/"
    exit
fi
readonly yaml_dir=$1
...
kubectl delete -f $yaml_dir/kip-nginx.yaml
```

scripts/run_eks_smoke_test.sh
```diff
@@ -0,0 +1,77 @@
...
+    kubectl get pods -A
+    kubectl get nodes
+    kubectl -n kip-smoke-tests describe pod -l statefulset.kubernetes.io/pod-name=kip-build-kip-provider-0
+    kubectl logs -n kip-smoke-tests -l statefulset.kubernetes.io/pod-name=kip-build-kip-provider-0r -c init-cert --tail=-1
+    kubectl logs -n kip-smoke-tests -l statefulset.kubernetes.io/pod-name=kip-build-kip-provider-0 -c kip --tail=-1
+}
+
+update_vk() {
+    local version="$(git describe)"
+    local patch_init="{\"spec\":{\"template\":{\"spec\":{\"initContainers\":[{\"image\":\"elotl/init-cert:$version\",\"name\":\"init-cert\"}]}}}}"
```
henryprecheur, line 35: That's a lot of of back-slahes escapes.

I'd use single quotes to avoid this:
```'{"spec":{"template":{"spec":{"initContainers":[{"image":"elotl/init-cert:","name":"init-cert"}]}}}}'```

File: scripts/run_eks_smoke_test.sh [hunk 3]
```diff
@@ -0,0 +1,77 @@
...
+    local patch_kip="{\"spec\":{\"template\":{\"spec\":{\"containers\":[{\"image\":\"elotl/kip:$version\",\"name\":\"kip\"}]}}}}"
+    kubectl patch -n kip-smoke-tests statefulset kip-build-kip-provider -p "$patch_init"
+    kubectl patch -n kip-smoke-tests statefulset kip-build-kip-provider -p "$patch_kip"
+    kubectl taint node kip-build-kip-provider-0 usage=kip-smoke-tests:NoSchedule --overwrite
+}
+
+run_smoke_test_1() {
+    kubectl apply -f $SCRIPT_DIR/eks-smoke-test/regular-node-nginx.yaml
+    local curlcmd="i=0; while [ \$i -lt 300 ]; do i=\$((i+1)); curl regular-nginx.kip-smoke-tests | grep 'Welcome to nginx' && exit 0; sleep 1; done; exit 1"
+    local waitcmd="phase=\"\"; echo \"Waiting for test results from pod\"; until [[ \$phase = Succeeded ]]; do sleep 1; phase=\$(kubectl get pod -n kip-smoke-tests test -ojsonpath=\"{.status.phase}\"); if [[ \$phase = Failed ]]; then echo \$phase; kubectl get pods -A; exit 1; fi; echo \$phase; done"
```
henryprecheur, line 45: curlcmd and waitcmd should be defined as readonly constants out of the function. I don't see any reason to have them be local variables.

scripts/run_eks_smoke_test.sh
```diff
@@ -0,0 +1,77 @@
...
+    kubectl run test --restart=Never --namespace=kip-smoke-tests --image=elotl/debug --command -- /bin/sh -c "$curlcmd"
+    timeout 420s bash -c "$waitcmd"
+    kubectl delete -n kip-smoke-tests pod test --ignore-not-found
+}
+
+run_smoke_test_2() {
+    kubectl apply -f $SCRIPT_DIR/eks-smoke-test/kip-nginx.yaml
+    local curlcmd="i=0; while [ \$i -lt 300 ]; do i=\$((i+1)); curl kip-nginx.kip-smoke-tests | grep 'Welcome to nginx' && exit 0; sleep 1; done; exit 1"
+    local waitcmd="phase=\"\"; echo \"Waiting for test results from pod\"; until [[ \$phase = Succeeded ]]; do sleep 1; phase=\$(kubectl get pod -n kip-smoke-tests test -ojsonpath=\"{.status.phase}\"); if [[ \$phase = Failed ]]; then echo \$phase; kubectl get pods -A; exit 1; fi; echo \$phase; done"
```
henryprecheur, line 54: same here

scripts/run_eks_smoke_test.sh
```diff
@@ -0,0 +1,77 @@
...
+    fetch_kubeconfig
+    update_vk
+    run_smoke_test_1
+    run_smoke_test_2
+    cleanup
+    trap - EXIT
+}
+
+# Run test_once() if running as a shell script, and return if sourced.
+(return 0 2>/dev/null) || test_once
```
henryprecheur, line 77: Why? Is that for debugging? I'd remove this.

scripts/run_eks_smoke_test.sh
```diff
@@ -0,0 +1,77 @@
...
+    kubectl delete -n kip-smoke-tests pod test --ignore-not-found
+}
+
+fetch_kubeconfig() {
+    echo "fetch kubeconfig"
+    aws eks update-kubeconfig --name elotl-ci-cd
+}
+
+test_once() {
+    set -euxo pipefail
```
henryprecheur, line 66: I think this should go to the beginning of the script.

henryprecheur (approved):

scripts/run_eks_smoke_test.sh
hidalgopl: My shellcheck is complaining about [this](https://github.com/koalaman/shellcheck/wiki/SC2016) if I use single quotes.

